PR #2, opened 2022-01-29
Efe_Adzhar_1_Terminal_HW1

EfeAdzhar:
general/1_terminal/Efe-Adzhar

### DIFF
--- a/EfeAdzhar/task2.sh
+++ b/EfeAdzhar/task2.sh
@@ -1,0 +1,5 @@
+#!/bin/zsh
+mkdir -p $1
+cd $1
+
+#To fix the issue, you hate to type mv ./mcd.sh nested/dir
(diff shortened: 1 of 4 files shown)


EfeAdzhar/task2.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/zsh
+mkdir -p $1
+cd $1
+
+#ANSWEAR:CD command works, but only inside the script because we launching it in a subprocess.
```
sae3023, line 5:
```suggestion
#ANSWER:CD command works, but only inside the script because we launching it in a subprocess.
```

sae3023 (dismissed):

EfeAdzhar/task2.sh
EfeAdzhar: Спасибо, не заметил

feketiko (approved):
Please, try not to name your commits as ".." in next homeworks
And avoid naming your branch as "EfeAdzhar" – use "Efe-Adzhar-1-hw" for example.

PR #4, opened 2022-01-29
Mykola_Tykhomyrov _1_Terminal_HW1

lilproger21:
Hello, it`s my home work


sae3023:
Let's remove the first task files from version control. 

Same goes for the `.odt` file. Add those to `.gitignore` or just don't `git add` them.

general/1_terminal/Mykola-Tykhomyrov/Home work 1/task_4.txt
```diff
@@ -0,0 +1 @@
+$ cat ./result.json | jq '.messages[].text'
```
sae3023, line 1: I'll accept this answer.

sae3023 (changes requested):

general/1_terminal/Mykola-Tykhomyrov/Home work 1/isWeekend(task_3).sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+today=$(date '+%a')
+if [ $today == 'Сб' ] || [ $today == 'Вс' ] || [ $today == 'Sat' ] || [ $today == 'Sun' ]
```
sae3023, line 4: Let's make it more portable.

Not all machines have Russian or English languages. Don't use string comparison in your final solution.

sae3023:
Also, don't forget the first task.

sae3023 (approved):

PR #5, opened 2022-01-29
Maksym_Omelianenko_1_Terminal_HW1

Omax0:


general/1_terminal/Maksym-Omelianenko/task4.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+set -e
```
sae3023, line 3: Good job on reading about `set -e`, this is widely used in production scrips.
I'd recommend using `set -ex`, so that the commands that are being executed are printed. More [here](https://www.peterbe.com/plog/set-ex).

general/1_terminal/Maksym-Omelianenko/task3.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
+set -e
```
sae3023, line 3: Good job on reading about `set -e`, this is widely used in production scrips.
I'd recommend using `set -ex`, so that the commands that are being executed are printed. More [here](https://www.peterbe.com/plog/set-ex).

sae3023 (approved):
Good job! If you wish to do an extra task, it's in one of the comments. Otherwise, feel free to merge, and don't forget to remove the branch.

sae3023:
Nooo!! You were supposed to merge.

PR #7, opened 2022-01-30
Max_Kovalchuk_Lecture1_HW

Maks90Kovalchuk:


general/1_terminal/Max-Kovalchuk/task1_homework.md
```diff
@@ -0,0 +1,19 @@
...
+
+# Output the names with "zoola"
+
+ls | grep "zoola"
+zoola_test
+zoola_api
+zoola-ac
+zoola
+
+
```
sae3023, line 18: Let's add a cleanup section that removes the files that this script creates.

general/1_terminal/Max-Kovalchuk/task4_homework.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+$1
+
+jq ".messages[].text" $1
```
sae3023, line 5: This is correct and the task is completed.

If you'd like an additional assignment, make it so that the messages that have links or mentions in them are rendered in the same way as the other messages.

Now, they are in an array.

sae3023 (changes requested):
The task solutions are accepted, nice work!

Let's remove the huge JSON file from version control and you can merge afterwards.

Also, I didn't mention it, but I'd really like for everyone to answer the questions in the same language that they were provided: English.

Same goes for the comments in the scripts, we just don't use Russian, even if every team member speaks it.

You don't have to rewrite the answer, but just keep it in mind for the future.

general/1_terminal/Max-Kovalchuk/task1_homework.md
sae3023: The zoolas are also to be gone.

sae3023 (approved):

PR #8, opened 2022-01-30
Oleh Levytskyi Terminal lesson_1

OlegLeva:


sae3023:
We don't need the chat export in version control. 

You can add it to `.gitignore` or just avoid `git add`ing it, your choice.

OlegLeva:
> We don't need the chat export in version control.
> 
> You can add it to `.gitignore` or just avoid `git add`ing it, your choice.

I added to .gitignore

general/1_terminal/Oleg-Levitskiy/task_1.txt
```diff
@@ -0,0 +1 @@
+exa -l *zoola*
```
sae3023, line 1: Solid choice of file system explorers.

I also prefer `exa` to `ls`, however, it's not installed on every machine by default. Let's use `ls`, so that it works everywhere.

general/1_terminal/Oleg-Levitskiy/task_3.sh
```diff
@@ -0,0 +1,13 @@
...
+day_week=$(date -d $todays_date +%A)
+day_month=$(date +%d)
+weekend='Saturday Sunday'
+
+if grep -q "$day_week" <<< "$weekend"
+then
+  echo Looks like $day_month is a weekend
+else
+  echo Looks like $day_month is a working day
+fi
```
sae3023, line 13: Let's make it so it works for our colleagues from Mexico and India.

Not everyone uses English as their terminal language. There exists a more portable solution.

general/1_terminal/Oleg-Levitskiy/task_4.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+cat zoola_chat_result.json | jq .messages[].text
```
sae3023, line 3: I'll accept this answer.

sae3023 (changes requested):
Some minor comments about portability, should take you 5 minutes to change.

The logic and the shape of the solution look good, but let's not compare against language-dependant strings.

Also, the `jq` assignment is accepted, and my comment there is optional.

sae3023:
Also, seems like the JSON is still in the working tree. Let's just remove it from git.

Finally, a minor comment: I haven't mentioned it anywhere, but I'd really like everyone to use English to answer the questions that were stated in English. You don't have to rewrite your answer, but just keep it in mind for the future

OlegLeva:
@sae3023 made all the changes you asked for. Or something else?

general/1_terminal/Oleg-Levitskiy/task_1.txt
```diff
@@ -0,0 +1 @@
+Ls -l *zoola*
```
sae3023, line 1:
```suggestion
ls -l *zoola*
```

sae3023 (approved):

sae3023 (approved):

PR #12, opened 2022-01-30
Danylo Kulbachinskiy hometask

magnetron-2000:


general/1_terminal/Danylo_Kulbachinskiy/script1.sh
```diff
@@ -0,0 +1,24 @@
+touch zoola1.txt
+touch zoola2.txt
+touch zoola3.txt
+touch zoom.txt
+touch hello.txt
+touch hizoola.txt
+
+set -e
```
sae3023, line 8: Good job on reading about `set -e`, this is a good thing to use in your scripts.

general/1_terminal/Danylo_Kulbachinskiy/script3.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+f="$(date "+%d")"
+arr=("субота" "неділя")
```
sae3023, line 3: Let's make the script more portable. Currently, this only works if your machine has Ukrainian language installed.

Unfortunately, it's not on every machine. And neither is English, so let's try to think of a solution that doesn't compare against strings.

general/1_terminal/Danylo_Kulbachinskiy/script4.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+
+
+jq '.messages[]' result.json | jq '.text' | tr -d 'a-z',"A-Z",'/"[]{}:\.','-_','&#'
```
sae3023, line 5: I like that you thought about making it prettier, but it mangled the output somewhat:
```
у кого есть конфиги вима  тоже скидывайте

  Я




  'ом пользоваться боюсь Там чтото страшное Плюс пока что меня спасал  +
 
 ```

I think it ate some of our text. The solution without `tr` would be accepted, but if you would like to try to make it prettier, you can do it, optionally.

sae3023 (changes requested):
Let's fix the portability issue.

Also, I didn't mention it during the lecture, but I think it would be more appropriate to use the English language to answer the questions that were stated in English. You don't have to change your current answer, it is accepted, but just keep it mind for the future.

sae3023 (approved):

PR #14, opened 2022-01-30
Shapkin Aleks HW Terminal1

Ashykay:
Shapkin Aleksandr, 1st_terminal
all solutions in a file "homework.md"

### DIFF
--- a/general/1_terminal/4_task.sh
+++ b/general/1_terminal/4_task.sh
@@ -1,0 +1,3 @@
+#!/bin/zsh
+
+jq '.messages[].text'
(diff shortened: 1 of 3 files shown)


general/1_terminal/4_task.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/zsh
+
+jq '.messages[].text'
```
sae3023, line 3: I'll accept this answer.

sae3023 (dismissed):

sae3023 (approved):

PR #20, opened 2022-01-31
Homework lectures 1,2 Anatolii Kavchenko

tol8901:
Hello, team.

This is the folder with homework about Lecture-1 Terminal, and processed with git according to the materials at Lecture-2 Git-1.
Could you check my homework, please?

Best regards,
Anatolii Kravchenko.

### DIFF
--- a/general/1_terminal/anatolii-kravchenko/task-1.sh
+++ b/general/1_terminal/anatolii-kravchenko/task-1.sh
@@ -9,3 +9,5 @@ for (( i = 0; i < $qtyOfFiles; i++ ))   # Loop for creating each file
 do
     touch "Zoola-$i.txt"
 done
+
+ls -l | grep Zoola                      # Output files where "Zoola" word is present in the filename
--- a/general/1_terminal/anatolii-kravchenko/task-1.sh
+++ b/general/1_terminal/anatolii-kravchenko/task-1.sh
@@ -3,7 +3,7 @@
 # This script is intended for create a bunch of files using touch. 
 # Output the names of those that have zoola in their names.
 
-qtyOfFiles=4                            # Set quantity of files that will be created
+qtyOfFiles=5                            # Set quantity of files that will be created
 
 for (( i = 0; i < $qtyOfFiles; i++ ))   # Loop for creating each file
 do
@@ -11,3 +11,8 @@ do
 done
 
 ls -l | grep Zoola                      # Output files where "Zoola" word is present in the filename
+
+for (( i = 0; i < $qtyOfFiles; i++ ))   # Loop for deleting each created Zoola- file
+do
+    rm "Zoola-$i.txt"
+done
--- a/general/1_terminal/anatolii-kravchenko/task-1.sh
+++ b/general/1_terminal/anatolii-kravchenko/task-1.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# This script is intended for create a bunch of files using touch. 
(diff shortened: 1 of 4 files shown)


general/1_terminal/anatolii-kravchenko/task-1.sh
```diff
@@ -0,0 +1,13 @@
...
+# Output the names of those that have zoola in their names.
+
+qtyOfFiles=4                            # Set quantity of files that will be created
+
+for (( i = 0; i < $qtyOfFiles; i++ ))   # Loop for creating each file
+do
+    touch "Zoola-$i.txt"
+done
+
+ls -l | grep Zoola                      # Output files where "Zoola" word is present in the filename
```
sae3023, line 13: Since we are creating files here, let's add a cleanup section, where we would delete everything that we have created.

general/1_terminal/anatolii-kravchenko/task-3.sh
```diff
@@ -0,0 +1,24 @@
...
+        ;;
+    2) echo "Looks like ${printDayNumber} is a working day" # Tuesday
+        ;;
+    3) echo "Looks like ${printDayNumber} is a working day" # Wednesday
+        ;;
+    4) echo "Looks like ${printDayNumber} is a working day" # Thursday
+        ;;
+    5) echo "Looks like ${printDayNumber} is a working day" # Friday
+        ;;
+    6) echo "Looks like ${printDayNumber} is a weekend"     # Saturday
```
sae3023, line 20: Nice of you to find the `case` statement that was not in the lecture.

However, I think we can find a better approach here. Look, for example, at how we repeat the same string 5 times. 
What you wrote here, essentially, is "for numbers between 1 and 5 do _this_, and for numbers 6 to 7 do _that_". We don't necessarily need to enumerate every number.

general/1_terminal/anatolii-kravchenko/task-4.txt
```diff
@@ -0,0 +1 @@
+jq '.messages | .[].text' ./result.json
```
sae3023, line 1: This is correct and the task is completed.

If you'd like an additional assignment, make it so that the messages that have links or mentions in them are rendered in the same way as the other messages.

Now, they are in an array.

sae3023 (changes requested):
Let's rewrite task number 3.

general/1_terminal/anatolii-kravchenko/task-1.sh
tol8901: I added this feature in last commit.

general/1_terminal/anatolii-kravchenko/task-3.sh
tol8901: I changed it to the IF statement, in last commit.

general/1_terminal/anatolii-kravchenko/task-4.txt
tol8901: I added this feature, in last commit.

general/1_terminal/anatolii-kravchenko/task-4.txt
```diff
@@ -0,0 +1 @@
+jq '.messages | .[].text | if type == "array" then [ .[] | if type == "object" then .text else . end] | add else . end | select(length > 0)' ./result.json
```
sae3023, line 1: Good job!

sae3023 (approved):

PR #37, opened 2022-02-03
updated location of the files

tumophiy:


general/1_terminal/Tima-Muravyov/README.md
```diff
@@ -1,10 +1,6 @@
 # homework 1
 
 
-![](images/task1.png)
```
feketiko, line 4: Do you need this change?

tumophiy: it will be better without image, saeryan said that and I agree cause repo will take less space

feketiko (approved):

PR #44, opened 2022-02-04
Move files to terminal folder

olnikiforov:


sae3023:
This PR only deletes the files. Are they already in a respective directory?

olnikiforov:
Most likely git shows this way, because after i writee the command "git add" it showed me that 5 files were deleted and a new folder was created

sae3023:
> Most likely git shows this way, because after i writee the command "git add" it showed me that 5 files were deleted and a new folder was created

We don't need an empty folder, we need a folder with your tasks. Please move the files to a directory corresponding to your name.

olnikiforov:
I made ctrl+X folder with homework to terminal folder. on remote it shows correctly

sae3023 (approved):

PR #99, opened 2022-07-21
Bogdana Maksiuta java lecture 01

danta-m:
The current pull request contain all home works to lecture 1.

### DIFF
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
@@ -1,0 +1,27 @@
+package com.zoolatech.lecture1.tasks._1;
+
+import java.util.Scanner;
+
+/**
+ * Write a program that accepts numbers of miles as a floating point number,
+ * converts it to kilometers and outputs the result.
+ *         a. Input: 2.5
+ *            Output: 4.0233603
+ */
+public class ConverterMiToKm {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        float miles = scanner.nextFloat(); //input number
+//        float miles = 2.5F; // for testing
+        if (miles > 0) {
+            converter(miles);
+        } else {
+            throw new IllegalArgumentException("Wrong number: " + miles);
+        }
+    }
+    public static void converter (float miles) {
+        final float METRIC_EQUIVALENT = (float) 1.609344;
+        float kilometer = miles * METRIC_EQUIVALENT;
+        System.out.println("From: " + miles + " mile \n" + "To: " + kilometer + " kilometer");
+    }
+}
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
@@ -1,0 +1,66 @@
+package com.zoolatech.lecture1.tasks._10;
+
+import java.util.Random;
+import java.util.Scanner;
+
+/**
+ * 10. Write a program that simulates the guess a number game.
+ * The program should generate a random number between 1 and N (N should be read from the input)
+ * and the user needs to guess it. If the user enters a number less than guessed,
+ * the program needs to output “Larger”. If the user enters a number greater than guessed -
+ * the program should output “Smaller”. If the user guesses the number correctly -
+ * the program should output “That’s it!” and finish.
+ * Input: 10
+ * Input: 7
+ * Output: Smaller
+ * Input: 3
+ * Output: Larger
+ * Input: 5
+ * Output: That’s it!
+ */
+public class GuessANumberGame {
+    static Scanner scanner = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        config();
+    }
+
+    public static void config () {
+        int maxLimit = 0;
+        while (maxLimit < 5 || maxLimit > 500) {
+            System.out.println("Start \nEnter the number \nFrom 5 to 500: ");
+            maxLimit = scanner.nextInt();
+            if (maxLimit > 5 && maxLimit < 500) {
+                game(maxLimit, randomNumber(maxLimit));
+            } else {
+                System.out.println("Invalid value. Try again");
+            }
+        }
+    }
+
+    public static int randomNumber (int maxLimit) {
+        Random randomGenerator = new Random();
+        return randomGenerator.nextInt(maxLimit);
+    }
+
+    public static void game (int maxLimit, int randomNumber) {
+        System.out.println("Pick a number between 1 and " + maxLimit + ". Enter your guess: ");
+        int inputNumber;
+        do {
+            inputNumber = scanner.nextInt();
+            if (inputNumber < 5) {
+                System.out.println("Nice try! But range is from 5 to " + maxLimit);
+            } else if (inputNumber > maxLimit) {
+                System.out.println("Nice try! But range is from 5 to " + maxLimit);
+            } else if (inputNumber < randomNumber) {
+                System.out.println("Larger");
+            } else if (inputNumber > randomNumber) {
+                System.out.println("Smaller");
+            } else {
+                System.out.println("That’s it!");
+                break;
+            }
+        } while (inputNumber != randomNumber);
+    }
+}
+
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_2/SpeedCalculator.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_2/SpeedCalculator.java
@@ -1,0 +1,42 @@
+package com.zoolatech.lecture1.tasks._2;
+
+import java.util.Scanner;
+
+/**
+ * Write a program that accepts a distance in meters and time in seconds and outputs
+ * a speed in meters/second and kilometers/hour. The result should be a floating point number.
+ *         a. Input: 80
+ *           10
+ *           Output: 8.0
+ * 	         28.8
+ */
+public class SpeedCalculator {
+    public static void main(String[] args) {
+        System.out.println("Введіть відстань: ");
+        Scanner scanner = new Scanner(System.in);
+        float distance = scanner.nextFloat();
+//        float distance = 80; //for testing
+        System.out.println("Введіть час: ");
+        Scanner scanner1 = new Scanner(System.in);
+        float time = scanner1.nextFloat();
+//        float time = 10; //for testing
+
+        if (distance >= 0 && time >= 0) {
+            speedMeterPerSecond(distance, time);
+            speedKilometerPerHour(distance, time);
+        } else {
+            throw new IllegalArgumentException("You entered a negative value");
+        }
+    }
+
+    public static void speedMeterPerSecond (float distance, float time) {
+        float speedMeterPerSecond = distance / time;
+        System.out.println("Speed is " + speedMeterPerSecond + " meters/second");
+    }
+
+    public static void speedKilometerPerHour (float distance, float time) {
+        final float METRIC_EQUIVALENT = 0.2777778f;
+        float speedKilometerPerHour = (distance / time) / METRIC_EQUIVALENT;
+        System.out.println("Speed is " + speedKilometerPerHour + " kilometers/hour");
+    }
+}
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_4/DefineDayOfTheWeek.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_4/DefineDayOfTheWeek.java
@@ -1,0 +1,38 @@
+package com.zoolatech.lecture1.tasks._4;
+
+import java.util.Scanner;
+
+/**
+ * Write a program that accepts a day of the week as an integer number (1 - Monday, 2 - Tuesday, …, 7 - Sunday)
+ * and output “Need to go to work…” if the day is a weekday or “Sleeping…” if it’s a weekend.
+ * Create two versions of the program: one should use the if-else statement, second - the conditional operator.
+ * a. Input: 3
+ * Output: “Need to go to work…”
+ * Explanation: 3 represents Wednesday, which is a weekday
+ * b. Input: 6
+ * Output: “Sleeping…”
+ * Explanation: 6 represents Saturday, which is a weekend
+ */
+public class DefineDayOfTheWeek {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter the day of the week: \nWhere \"1\" is Monday \"7\" is Sunday");
+        int numberOfTheDay = scanner.nextInt();
+        ifElseDefiner(numberOfTheDay);
+        System.out.println(conditionalOperatorDefiner(numberOfTheDay));
+    }
+
+    public static void ifElseDefiner(int numberOfTheDay) {
+        if (numberOfTheDay >= 1 && numberOfTheDay <= 5) {
+            System.out.println("Need to go to work...");
+        } else if (numberOfTheDay >= 6 && numberOfTheDay <= 7) {
+            System.out.println("Sleeping…");
+        } else {
+            throw new IllegalArgumentException("Wrong number " + numberOfTheDay);
+        }
+    }
+
+    public static String conditionalOperatorDefiner(int numberOfTheDay) {
+        return numberOfTheDay >= 1 && numberOfTheDay <= 5 ? "Need to go to work..." : "Sleeping…";
+    }
+}
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_5/LargestNumber.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_5/LargestNumber.java
@@ -1,0 +1,33 @@
+package com.zoolatech.lecture1.tasks._5;
+
+import java.util.Scanner;
+
+/**
+ * 5. Write a program that accepts two numbers and outputs the larger number.
+ * If numbers are equal - output “Numbers are equal”.
+ *         a. Input: 2
+ *            3
+ *            Output: 3
+ *         b. Input: 2
+ *            2
+ *            Output: “Numbers are equal”
+ */
+public class LargestNumber {
+    public static void main(String[] args) {
+        System.out.println("Enter 2 numbers: ");
+        Scanner scanner = new Scanner(System.in);
+        int firstNumber = scanner.nextInt();
+        int secondNumber = scanner.nextInt();
+
+        compareNumbers(firstNumber, secondNumber);
+    }
+    public static void compareNumbers (int firstNumber, int secondNumber) {
+        if (firstNumber > secondNumber) {
+            System.out.println("The largest number is: " + firstNumber);
+        } else if (firstNumber == secondNumber) {
+            System.out.println("The numbers are equal");
+        } else {
+            System.out.println("The largest number is: " + secondNumber);
+        }
+    }
+}
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
@@ -1,0 +1,97 @@
+package com.zoolatech.lecture1.tasks._6;
+
+import java.util.Scanner;
+
+/**
+ * 6. Write a program that accepts two numbers and a symbol of an operation (as a character)
+ * and outputs the result of the selected operation. Possible characters for the operation:
+ * ‘+’, ‘-’, ‘*’, ‘/’, ‘%’. The program should work only with integer values.
+ * Create two versions of the program: one should use the if-else statement, second - the switch statement.
+ *         a. Input: 7
+ *            5
+ *            +
+ * Output: 12
+ *         b. Input: 7
+ *           5
+ *           %
+ * 		Output: 2
+ */
+public class Calculator {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter first number:");
+        int firstNumber = scanner.nextInt();
+        System.out.println("Enter second number:");
+        int secondNumber = scanner.nextInt();
+        System.out.println("Add arithmetic operation:");
+        char symbol = scanner.next().charAt(0);
+
+        ifElseCalculator(firstNumber, secondNumber, symbol);
+        switchCaseCalculator(firstNumber, secondNumber, symbol);
+    }
+
+    public static void ifElseCalculator(int firstNumber, int secondNumber, char symbol) {
+        if (symbol == '+') {
+            int result = firstNumber + secondNumber;
+            System.out.println(firstNumber + " + " + secondNumber + " = " + result);
+        }
+        if (symbol == '-') {
+            int result = firstNumber - secondNumber;
+            System.out.println(firstNumber + " - " + secondNumber + " = " + result);
+        }
+        if (symbol == '*') {
+            int result = firstNumber * secondNumber;
+            System.out.println(firstNumber + " * " + secondNumber + " = " + result);
+        }
+        if (symbol == '/') {
+            if (secondNumber == 0) {
+                System.out.println("Cannot divide by 0");
+            } else {
+                int result = firstNumber / secondNumber;
+                System.out.println(firstNumber + " / " + secondNumber + " = " + result);
+            }
+        }
+        if (symbol == '%') {
+            if (secondNumber == 0) {
+                System.out.println("Cannot divide by 0");
+            } else {
+                int result = firstNumber % secondNumber;
+                System.out.println(firstNumber + " % " + secondNumber + " = " + result);
+            }
+        }
+    }
+
+    public static void switchCaseCalculator(int firstNumber, int secondNumber, char symbol) {
+        switch (symbol) {
+            case '+' -> {
+                int result = firstNumber + secondNumber;
+                System.out.println(firstNumber + " + " + secondNumber + " = " + result);
+            }
+            case '-' -> {
+                int result = firstNumber - secondNumber;
+                System.out.println(firstNumber + " - " + secondNumber + " = " + result);
+            }
+            case '*' -> {
+                int result = firstNumber * secondNumber;
+                System.out.println(firstNumber + " * " + secondNumber + " = " + result);
+            }
+            case '/' -> {
+                if (secondNumber == 0) {
+                    System.out.println("Cannot divide by 0");
+                } else {
+                    int result = firstNumber / secondNumber;
+                    System.out.println(firstNumber + " / " + secondNumber + " = " + result);
+                }
+            }
+            case '%' -> {
+                if (secondNumber == 0) {
+                    System.out.println("Cannot divide by 0");
+                } else {
+                    int result = firstNumber % secondNumber;
+                    System.out.println(firstNumber + " % " + secondNumber + " = " + result);
+                }
+            }
+        }
+    }
+}
+
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_7/OddNumbers.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_7/OddNumbers.java
@@ -1,0 +1,57 @@
+package com.zoolatech.lecture1.tasks._7;
+
+import java.util.Scanner;
+
+/**
+ * 7. Write a program that prints all odd numbers from N to 1. N should be read from the input.
+ * Every number should be printed on the new line. Create two versions of the program:
+ * one should use the for loop, second - the while loop.
+ *         a. Input: 5
+ *            Output: 5
+ *                    3
+ *                    1
+ */
+public class OddNumbers {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter a number: ");
+        int number = scanner.nextInt();
+
+        forLoopAllOddNumbers(number);
+        whileLoopAllOddNumbers(number);
+    }
+
+    public static void forLoopAllOddNumbers(int number) {
+        System.out.println("For loop: ");
+        if (number >= 0) {
+            for (int i = number; i > 0; i--) {
+                if (i % 2 == 1) {
+                    System.out.println(i);
+                }
+            }
+        } else {
+            for (int i = number; i < 1; i++) {
+                if (i % 2 == -1) {
+                    System.out.println(i);
+                }
+            }
+        }
+    }
+
+    public static void whileLoopAllOddNumbers(int number) {
+        System.out.println("While loop: ");
+        while (number > 0) {
+            if (number % 2 == 1) {
+                System.out.println(number);
+            }
+            number--;
+        }
+        while (number < 0) {
+            if (number % 2 == -1) {
+                System.out.println(number);
+            }
+            number++;
+        }
+    }
+}
+
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
@@ -1,0 +1,68 @@
+package com.zoolatech.lecture1.tasks._8;
+
+import java.util.Scanner;
+
+/**
+ * 8. Write a program that prints all numbers from 1 to N, that are divisible by 2, 3 or by both.
+ * If a number is divisible by only 2 - print it and add in parentheses “(by 2)”, if only by 3 - “(by 3)”,
+ * if by both numbers - “(by 2 and 3)”. Every number should be printed on the new line.
+ * Create two versions of the program: one should use the for loop, second - the while loop.
+ * a. Input: 12
+ * Output: 2 (by 2)
+ *         3 (by 3)
+ *         4 (by 2)
+ *         6 (by 2 and 3)
+ *         8 (by 2)
+ *         9 (by 3)
+ *         10 (by 2)
+ *         12 (by 2 and 3)
+ */
+public class CheckDivideIsPossible {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter a number: ");
+        int number = scanner.nextInt();
+        divisionForLoop(number);
+        divisionWhileLoop(number);
+    }
+
+    public static void divisionForLoop(int number) {
+        System.out.println("For loop: ");
+        if (number >= 0) {
+            for (int i = 1; i <= number; i++) {
+                divisionBy2By3(i);
+            }
+        } else {
+            for (int i = number; i < 0; i++) { // for negative numbers
+                divisionBy2By3(i);
+            }
+        }
+    }
+
+    public static void divisionBy2By3(int i) {
+        if (i % 2 == 0 && i % 3 == 0) {
+            System.out.println(i + " (by 2 and 3)");
+        } else if (i % 2 == 0) {
+            System.out.println(i + " (by 2)");
+        } else if (i % 3 == 0) {
+            System.out.println(i + " (by 3)");
+        }
+    }
+
+    public static void divisionWhileLoop(int number) {
+        System.out.println("\nWhile loop: ");
+        int startNumber = 1;
+
+        if (number >= 0) {
+            while (startNumber <= number) {
+                divisionBy2By3(startNumber);
+                startNumber++;
+            }
+        } else {
+            while (number < 0) { // for negative numbers
+                divisionBy2By3(number);
+                number++;
+            }
+        }
+    }
+}
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_9/Reverse.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_9/Reverse.java
@@ -1,0 +1,45 @@
+package com.zoolatech.lecture1.tasks._9;
+
+import java.util.Scanner;
+
+/**
+ * 9. Write a program that accepts an integer value and outputs all its digits in the reverse order on a new line.
+ a. Input: 12345
+ *      Output: 5
+ *              4
+ *              3
+ *              2
+ *              1
+ */
+public class Reverse {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter a number: ");
+        int number = scanner.nextInt();
+
+        reverseNumberWhileLoop(number);
+        reverseNumberForLoop (number);
+    }
+
+    public static void reverseNumberWhileLoop (int number) {
+        System.out.println("While loop: ");
+        int numberForReverse = 0;
+        while (number != 0) {
+            int lastNumber = number % 10;
+            numberForReverse = numberForReverse * 10 + lastNumber;
+            number = number/10;
+            System.out.println(lastNumber);
+        }
+    }
+
+    public static void reverseNumberForLoop (int number) {
+        System.out.println("For loop: ");
+        int numberForReverse = 0;
+        for (int i = number; i != 0 ; i /= 10) {
+            int lastNumber = i % 10;
+            System.out.println(lastNumber);
+            numberForReverse = numberForReverse * 10 + lastNumber;
+        }
+    }
+}
+
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .idea/
 **/.DS_Store
+out/
--- a/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_3/FractionalPartExtractor.java
+++ b/java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_3/FractionalPartExtractor.java
@@ -1,0 +1,31 @@
+package com.zoolatech.lecture1.tasks._3;
+
+import java.util.Scanner;
+
+/**
+ *     3. Write a program that accepts a double number and outputs only its fractional part
+ *     (all digits after the decimal point). Note: it’s fine if the result will have a rounding error
+ *     (0.8499999999999 instead of 0.85).
+ *             a. Input: 1.23
+ *                Output: 0.22999999999999998
+ */
+public class FractionalPartExtractor {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Enter a number:");
+
+        if (scanner.hasNextDouble()) {
+            double inputNumber = scanner.nextDouble();
+            fractionalPart(inputNumber);
+        } else {
+            throw new IllegalArgumentException("Unknown number");
+        }
+    }
+
+    public static void fractionalPart (double inputNumber) {
+        int inputNumberInteger = (int) inputNumber;
+        double fractionalPartInputNumber = inputNumber - inputNumberInteger;
+        System.out.println(fractionalPartInputNumber);
+    }
+}
+


java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
```diff
@@ -0,0 +1,27 @@
...
+        float miles = scanner.nextFloat(); //input number
+//        float miles = 2.5F; // for testing
+        if (miles > 0) {
+            converter(miles);
+        } else {
+            throw new IllegalArgumentException("Wrong number: " + miles);
+        }
+    }
+    public static void converter (float miles) {
+        final float METRIC_EQUIVALENT = (float) 1.609344;
```
tromanyuk, line 23: Instead of casting, you can simply add to the literal the letter `F`.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_2/SpeedCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+ * 	         28.8
+ */
+public class SpeedCalculator {
+    public static void main(String[] args) {
+        System.out.println("Введіть відстань: ");
+        Scanner scanner = new Scanner(System.in);
+        float distance = scanner.nextFloat();
+//        float distance = 80; //for testing
+        System.out.println("Введіть час: ");
+        Scanner scanner1 = new Scanner(System.in);
```
tromanyuk, line 20: It's possible to use a single instance of the scanner class, you don't need to create a second one.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
```diff
@@ -0,0 +1,27 @@
...
+ * Write a program that accepts numbers of miles as a floating point number,
+ * converts it to kilometers and outputs the result.
+ *         a. Input: 2.5
+ *            Output: 4.0233603
+ */
+public class ConverterMiToKm {
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        float miles = scanner.nextFloat(); //input number
+//        float miles = 2.5F; // for testing
```
tromanyuk, line 15: I understand that this was added for testing, but pretend that it's a real code - it mustn't contain such statements

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_2/SpeedCalculator.java
```diff
@@ -0,0 +1,42 @@
...
+        float time = scanner1.nextFloat();
+//        float time = 10; //for testing
+
+        if (distance >= 0 && time >= 0) {
+            speedMeterPerSecond(distance, time);
+            speedKilometerPerHour(distance, time);
+        } else {
+            throw new IllegalArgumentException("You entered a negative value");
```
tromanyuk, line 28: nice

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_4/DefineDayOfTheWeek.java
```diff
@@ -0,0 +1,38 @@
...
+            System.out.println("Need to go to work...");
+        } else if (numberOfTheDay >= 6 && numberOfTheDay <= 7) {
+            System.out.println("Sleeping…");
+        } else {
+            throw new IllegalArgumentException("Wrong number " + numberOfTheDay);
+        }
+    }
+
+    public static String conditionalOperatorDefiner(int numberOfTheDay) {
+        return numberOfTheDay >= 1 && numberOfTheDay <= 5 ? "Need to go to work..." : "Sleeping…";
```
tromanyuk, line 36: What if the value is out of the allowed range?

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_5/LargestNumber.java
```diff
@@ -0,0 +1,33 @@
...
+    public static void main(String[] args) {
+        System.out.println("Enter 2 numbers: ");
+        Scanner scanner = new Scanner(System.in);
+        int firstNumber = scanner.nextInt();
+        int secondNumber = scanner.nextInt();
+
+        compareNumbers(firstNumber, secondNumber);
+    }
+    public static void compareNumbers (int firstNumber, int secondNumber) {
+        if (firstNumber > secondNumber) {
```
tromanyuk, line 25: You have almost the same statement if numbers are not equal. Try to optimize it.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
```diff
@@ -0,0 +1,97 @@
...
+        System.out.println("Enter second number:");
+        int secondNumber = scanner.nextInt();
+        System.out.println("Add arithmetic operation:");
+        char symbol = scanner.next().charAt(0);
+
+        ifElseCalculator(firstNumber, secondNumber, symbol);
+        switchCaseCalculator(firstNumber, secondNumber, symbol);
+    }
+
+    public static void ifElseCalculator(int firstNumber, int secondNumber, char symbol) {
```
tromanyuk, line 33: Try to extract common statements out of if/else blocks. Same for the switch statement

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_7/OddNumbers.java
```diff
@@ -0,0 +1,57 @@
...
+    public static void forLoopAllOddNumbers(int number) {
+        System.out.println("For loop: ");
+        if (number >= 0) {
+            for (int i = number; i > 0; i--) {
+                if (i % 2 == 1) {
+                    System.out.println(i);
+                }
+            }
+        } else {
+            for (int i = number; i < 1; i++) {
```
tromanyuk, line 33: Interesting, I haven't thought about this case. But, let's assume the number is positive.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public static void divisionForLoop(int number) {
+        System.out.println("For loop: ");
+        if (number >= 0) {
+            for (int i = 1; i <= number; i++) {
+                divisionBy2By3(i);
+            }
+        } else {
+            for (int i = number; i < 0; i++) { // for negative numbers
```
tromanyuk, line 36: `from 1 to N`

this excludes negative numbers

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
```diff
@@ -0,0 +1,68 @@
...
+                divisionBy2By3(i);
+            }
+        }
+    }
+
+    public static void divisionBy2By3(int i) {
```
tromanyuk, line 42: Nice idea to extract common functionality to a different method. Just note, that such methods are usually defined after the last usage (if that's possible).

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_9/Reverse.java
```diff
@@ -0,0 +1,45 @@
...
+        System.out.println("Enter a number: ");
+        int number = scanner.nextInt();
+
+        reverseNumberWhileLoop(number);
+        reverseNumberForLoop (number);
+    }
+
+    public static void reverseNumberWhileLoop (int number) {
+        System.out.println("While loop: ");
+        int numberForReverse = 0;
```
tromanyuk, line 26: How do we use the `numberForReverse`? Same question for the second loop.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
```diff
@@ -0,0 +1,66 @@
...
+public class GuessANumberGame {
+    static Scanner scanner = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        config();
+    }
+
+    public static void config () {
+        int maxLimit = 0;
+        while (maxLimit < 5 || maxLimit > 500) {
```
tromanyuk, line 30: Why have you chosen such boundaries for the max value - 5 and 500? The task description doesn't say anything regarding the boundaries of the max value.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
```diff
@@ -0,0 +1,66 @@
...
+            inputNumber = scanner.nextInt();
+            if (inputNumber < 5) {
+                System.out.println("Nice try! But range is from 5 to " + maxLimit);
+            } else if (inputNumber > maxLimit) {
+                System.out.println("Nice try! But range is from 5 to " + maxLimit);
+            } else if (inputNumber < randomNumber) {
+                System.out.println("Larger");
+            } else if (inputNumber > randomNumber) {
+                System.out.println("Smaller");
+            } else {
```
tromanyuk, line 59: One more approach would be to remove the else block and move the println statement out of the loop, since the exit condition will be checked by the loop

But, this will work as well.

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Calculator.java
```diff
@@ -0,0 +1,70 @@
+package com.zoolatech.lecture2.tasks._1;
+
+import java.util.Scanner;
+
+public class Calculator {
+    Scanner scanner = new Scanner(System.in);
+
+    public void processOfCalculation() {
```
tromanyuk, line 8: Methods processOfCalculation, inputNumber, inputSymbol should be a part of the calculator API (see task's description). They can be defined in the Main class, and later you can use the calculator instance to perform a specific operation.

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Calculator.java
```diff
@@ -0,0 +1,70 @@
...
+
+    public double inputNumber() {
+        System.out.println("Please enter a number : ");
+        double number;
+        if (scanner.hasNextDouble()) {
+            number = scanner.nextDouble();
+        } else {
+            System.out.println("Invalid value. Try another number : ");
+            scanner.next();
+            number = inputNumber();
```
tromanyuk, line 30: Interesting idea with the recursion. However, prefer to use a loop over the recursion since the later one has a limited number of operations (it's hard to receive it practically, but still) and uses more memory

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Main.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public static void main(String[] args) {
+        Calculator calculation = new Calculator();
+        calculation.processOfCalculation();
+
+        Scanner scanner = new Scanner(System.in);
+
+        boolean work = true;
+        while (work) {
+            System.out.println("""
```
tromanyuk, line 23: great

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Main.java
```diff
@@ -0,0 +1,38 @@
...
+                                Next calculation?
+                                'Y' - YES
+                                'N' - NO
+                                """);
+            String choice = scanner.next();
+            if ("Y".equals(choice)) {
```
tromanyuk, line 29: What if a value is neither `Y` or `N`?

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Main.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+public class Main {
+
+    public static void main(String[] args) {
+        Calculator calculation = new Calculator();
+        calculation.processOfCalculation();
+
+        Scanner scanner = new Scanner(System.in);
+
+        boolean work = true;
```
tromanyuk, line 21: probably, you could've remove the variable `work` if you'd use a do-while loop with the variable `choice` inside of the loop's condition

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/VacuumCleaner.java
```diff
@@ -0,0 +1,23 @@
...
+ * The north-west corner of a room is represented by coordinates (0,0); north-east - (W-1, 0);
+ * south-east - (W-1, H-1); south-west - (0, H-1).
+ * By default, a robot should be placed in the north-west corner facing south direction.
+ * However, it should be also possible to give any coordinates and direction as the initial robot location.
+ * Robot must not be able to get outside of a room - in such case the robot should move as much as possible
+ * to the room border, print a warning message and stop. All next moves in the same direction should be prevented.
+ */
+public class VacuumCleaner {
+
+    public static void main(String[] args) {
```
tromanyuk, line 17: Please carefully read the task description: you need to have a class that provides following methods: 
- move()
- move(int tiles)
- turnLeft()
- turnRight()
- currentLocation()
- currentDirection()
- constructors to properly create an object of this class

After that, you can add the Main class where you can create an instance of this class and accept the input from the user. However, this part is optional. The main goal is to define the class.

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Drum.java
```diff
@@ -0,0 +1,16 @@
+package com.zoolatech.lecture3.tasks._1;
+
+public class Drum extends MusicalInstrument {
+    public Drum(String name, String sound) {
+        super(name, sound);
+    }
+
+    void play() {
+        super.play();
```
tromanyuk, line 9: You don't need to call the parent's method or even define the method `play` in the children classes

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Drum.java
```diff
@@ -0,0 +1,16 @@
+package com.zoolatech.lecture3.tasks._1;
+
+public class Drum extends MusicalInstrument {
```
tromanyuk, line 3: Do we really need to create a class for each instrument? Do they provide some additional functionality?

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Musicians.java
```diff
@@ -0,0 +1,37 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.Objects;
+
+public class Musicians {
```
tromanyuk, line 5: `Musician`, singular form

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Orchestra.java
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    public void remove(Musicians musicians) {
+        listOfMusicians.remove(musicians);
+        System.out.println(musicians + " left first group");
+    }
+
+    public void playAll() {
+        for (Musicians m : listOfMusicians) {
+            m.instrument.play();
```
tromanyuk, line 20: It would be better to add some method to a musician that will play music instead of accessing a field. It would also allow you to make this field private.

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Main.java
```diff
@@ -0,0 +1,45 @@
...
+        Musicians trumpeter = new Musicians("Mike", trumpet);
+        Musicians drummer = new Musicians("Peter", drum);
+
+        System.out.println("The guitarist and the drummer are the same person? " + guitarist.equals(drummer));
+        guitarist.getNameAndInstrument(guitarist);
+        trumpeter.getNameAndInstrument(trumpeter);
+        drummer.getNameAndInstrument(drummer);
+
+
+        Orchestra firstGroup = new Orchestra();
```
tromanyuk, line 30: Why `group` if the class is called `Orchestra`? 😄

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/Main.java
```diff
@@ -0,0 +1,37 @@
...
+ * “validating that fieldName is less than 10” and return true from a smaller validator.
+ * In the main method, create a user account and user account validator objects.
+ * The validator needs to contain all validation rules that can be applied to user account objects.
+ */
+public class Main {
+    public static void main(String[] args) {
+        UserAccount userAccount = new UserAccount(1, "Masha", "Lup",
+                "Ukraine", "boo@i.ua", "00-00-01", 3, 5);
+
+        UserAccountValidator validation = new UserAccountValidator();
```
tromanyuk, line 24: validation -> validator

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,140 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public class UserAccountValidator implements Validator {
+    IdUserValidator idUserValidator = new IdUserValidator();
```
tromanyuk, line 4: `The validator needs to use a list of smaller validators to perform a validation of specific account fields.`

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,140 @@
...
+            System.out.println("Access granted");
+        } else {
+            System.out.println("Access denied");
+        }
+        return isValid;
+    }
+
+}
+
+class IdUserValidator implements Validator {
```
tromanyuk, line 30: `The smaller validators need to accept a field name during the creation and apply a specific validation rule to this field.`

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,140 @@
...
+    public boolean isCorrect(UserAccount userAccount) {
+        System.out.println("Validating that 'id' is correct");
+        return userAccount.getUserId() > 0;
```
tromanyuk, line 33: `You don’t need to perform actual validation, it’s enough to print smth like “validating that fieldName is less than 10” and return true from a smaller validator.`

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public boolean isValid(UserAccount userAccount) {
+        return isCorrect(userAccount);
+    }
+}
+
+class NameValidator implements Validator {
```
tromanyuk, line 42: You don't need to define a new validator class for each field - you need to create a validator for each rule type (see task's description) and then pass a field name to an object of this validator.

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
```diff
@@ -0,0 +1,45 @@
+package com.zoolatech.lecture3.tasks._3;
+
+import java.util.Objects;
+
+public non-sealed class Circle implements Shape {
+    final double PI = 3.1415926536;
```
tromanyuk, line 6: I think it can be static

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+    int radius;
```
tromanyuk, line 7: prefer to use private fields

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public Circle(int radius) {
+        this.radius = radius;
+    }
+
+    @Override
+    public void findPerimeter() {
+        int perimeter = (int) (radius * 2 * PI);
```
tromanyuk, line 15: why int and not a floating-point number?

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public void findArea() {
+        int area = (int) (Math.pow(radius, 2) * PI);
+        System.out.println("The area of the circle is " + area);
+    }
+
+    @Override
+    public String toString() {
+        return "Circle: " +
+                "PI = " + PI +
```
tromanyuk, line 28: `PI` will be the same for all classes, you should be only interested in radius

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Rectangle.java
```diff
@@ -0,0 +1,45 @@
+package com.zoolatech.lecture3.tasks._3;
+
+import java.util.Objects;
+
+public non-sealed class Rectangle implements Shape {
+    int firstSide;
```
tromanyuk, line 6: same: you need to use private fields to enforce encapsulation

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Main.java
```diff
@@ -0,0 +1,42 @@
...
+        Triangle triangle3 = new Triangle(3, 4, 6);
+
+        circle.findArea();
+        circle.findPerimeter();
+        rectangle.findArea();
+        rectangle.findPerimeter();
+        triangle.findArea();
+        triangle.findPerimeter();
+
+        System.out.println("\nThe radius of a circle is: " + circle.radius);
```
tromanyuk, line 29: `Then display the radius of a circle and lengths of a triangle and rectangle sides for objects stored in the list.`

Same: you need to call specific methods of every object from the loop

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
```diff
@@ -0,0 +1,68 @@
...
+    String fileName;
+
+    public Reporter(String headerData, String fileData, String fileName) {
+        this.headerData = headerData;
+        this.fileData = fileData;
+        this.fileName = fileName;
+    }
+
+    void generate() {
+        System.out.println("\nWhat do you want? \nPress '1' for create a file \nPress '2' for add header data" +
```
tromanyuk, line 17: The method generate doesn't need to accept user input. It needs to perform all tasks (create a new file, add header data in an appropriate format (CSV, PDF, DOC), add file data in an appropriate format and close a file) when it's called.

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
```diff
@@ -0,0 +1,68 @@
+package com.zoolatech.lecture3.tasks._4;
+
+import java.util.Scanner;
+
+public abstract class Reporter {
```
tromanyuk, line 5: As far as I can see, right now all 3 subclasses have the behaviour. But, they must do different things (e.g., print "“Adding file details in CSV format”) which depend on type of documents they work with.

tromanyuk:
Nice work! Please carefully read the task description. You need to redo some tasks from the second and third homework, and fix smaller comments for other tasks.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_2/SpeedCalculator.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_1/ConverterMiToKm.java
danta-m: ok. understand. fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_4/DefineDayOfTheWeek.java
danta-m: added one more statement

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_5/LargestNumber.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_7/OddNumbers.java
danta-m: don't quite understand this condition. It's means that we don't need validation for the negative number?

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_8/CheckDivideIsPossible.java
danta-m: will try to follow this rule

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_9/Reverse.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
danta-m: it's was easy for me to test my code. fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_10/GuessANumberGame.java
```diff
@@ -0,0 +1,66 @@
...
+                game(maxLimit, randomNumber(maxLimit));
+            } else {
+                System.out.println("Invalid value. Try again");
+            }
+        }
+    }
+
+    public static int randomNumber (int maxLimit) {
+        Random randomGenerator = new Random();
+        return randomGenerator.nextInt(maxLimit);
```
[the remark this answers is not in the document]
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Calculator.java
```diff
@@ -0,0 +1,70 @@
...
+        System.out.println("Please enter a symbol of arithmetic operation : ");
+        char symbol = scanner.next().charAt(0);
+        if (symbol != '/' && symbol != '+' && symbol != '-' && symbol != '*') {
+            System.out.println("Invalid value");
+            symbol = inputSymbol();
+        }
+        return symbol;
+    }
+
+    public void addition(double firstNumber, double secondNumber) {
```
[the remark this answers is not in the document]
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Calculator.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Main.java
danta-m: rewrited code

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/VacuumCleaner.java
danta-m: rewrited code. Not clear to understand the main goal.

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_1/Main.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Drum.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Drum.java
danta-m: no, we don't need it. fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Musicians.java
danta-m: ops, fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Orchestra.java
```diff
@@ -0,0 +1,23 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.ArrayList;
+
+public class Orchestra {
+    ArrayList<Musicians> listOfMusicians = new ArrayList<>();
```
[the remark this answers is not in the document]
danta-m: done

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Orchestra.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Main.java
danta-m: my bad. the orchestra consists of groups, but in our case it's actually an orchestra

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccount.java
```diff
@@ -0,0 +1,58 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import lombok.Getter;
```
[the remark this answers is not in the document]
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/Main.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
danta-m: hope i fixed it

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
danta-m: we don't validate "id"? Is there a difference in the implementation of the tasks when all the classes are in 1 file and each file is in different ones? or is it visually better?

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
danta-m: without reason. fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
danta-m: fixed. Missed this moment

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Rectangle.java
danta-m: done

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Main.java
```diff
@@ -0,0 +1,42 @@
...
+ * In the main method, create a list of three objects and display their perimeter and area.
+ * Then display the radius of a circle and lengths of a triangle and rectangle sides for objects stored in the list.
+ * Prevent other classes from extension from the Shape interface.
+ * It should be also possible to check if two object references of these types are the same and
+ * to print the inner state of an object (like radius or width) to the console output when
+ * an object is passed to the System.out.println method.
+ */
+public class Main {
+    public static void main(String[] args) {
+        Circle circle = new Circle(5);
```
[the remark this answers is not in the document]
danta-m: done

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Main.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
danta-m: fixed

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
danta-m: didn't understand the task. I hope that it is correct now

danta-m:
The last time and my last commit is "to remove /out from /gitignore" . And after pushing I have a lot of unnecessary files in my request. I have no idea what to do to solve this problem and remove all "/out' files from my request

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
```diff
@@ -0,0 +1,87 @@
...
+
+        System.out.println(firstNumber + " " + symbol + " " + secondNumber + " = " +
+                ifElseCalculator(firstNumber, secondNumber, symbol));
+        System.out.println(firstNumber + " " + symbol + " " + secondNumber + " = " +
+                switchCaseCalculator(firstNumber, secondNumber, symbol));
+    }
+
+    public static int ifElseCalculator(int firstNumber, int secondNumber, char symbol) {
+        int result = 0;
+        if (symbol == '+') {
```
tromanyuk, line 37: Such approach will work as well but it's better to use the if-else syntax: current solution will examine every condition (symbol == '+', symbol == '-', and so on). If-else after it finds the match won't check next conditions. This reduces a chance of a possible bug.

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_6/Calculator.java
```diff
@@ -0,0 +1,87 @@
...
+            } else {
+                result = firstNumber % secondNumber;
+            }
+        }
+        return result;
+    }
+
+    public static int switchCaseCalculator(int firstNumber, int secondNumber, char symbol) {
+        int result = 0;
+        switch (symbol) {
```
tromanyuk, line 65: you can use the switch expression here: use the result of the switch expressions as a value for a variable

java/bogdana_maksiuta/src/com/zoolatech/lecture1/tasks/_7/OddNumbers.java
tromanyuk: yes, we assume all numbers are positive

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+ * move()
+ * move(int tiles)
+ * turnLeft()
+ * turnRight()
+ * currentLocation()
+ * currentDirection()
+ * constructors to properly create an object of this class
+ */
+public class Robot {
+    private int h = 0;
```
tromanyuk, line 15: `0` is the default value for int variables, so, usually, such assignments can be omitted

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+    public int getHeightRoom() {
+        return heightRoom;
+    }
+    public void setHeightRoom(int heightRoom) {
+        this.heightRoom = heightRoom;
+    }
+
+    public void move() {
+        switch (getDirection()) {
```
tromanyuk, line 52: same: you can access to instance variables **inside of a class** directly, without a getter method

tromanyuk, line 52: the code of the method `move()` can be replaced with the call to the method `move(int tiles)`: 

```
move(1);
```

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+            case "south" -> {
+                h++;
+                southDirection();
+            }
+        }
+    }
+
+    public void move(int tiles) {
+        switch (getDirection()) {
+            case "east" -> {
```
tromanyuk, line 74: I think it might be better to define directions as class constants (private static final String EAST = "east") or as enum Direction

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+    public void currentDirection() {
+        System.out.println("Direction: " + getDirection());
+    }
+
+    public void southDirection() {
+        if (h < (heightRoom - 1)) {
+            System.out.println("My direction: SOUTH");
+            setDirection("south");
+        } else if (h == (heightRoom - 1) && w < heightRoom / 2 && w >= 0) {
+            System.out.println("Hm. It's border. I need to go east");
```
tromanyuk, line 143: nice idea, but the task description was easier: you could just print that you stuck at the border

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+            }
+        }
+    }
+
+    public void move(int tiles) {
+        switch (getDirection()) {
+            case "east" -> {
+                if ((getW() + tiles) < getWidthRoom()) {
+                    w += tiles;
+                    eastDirection();
```
tromanyuk, line 77: We have checked that we should be able to move to selected number of tiles; do we need to call this method? it will just set the direction as `east` (which is already `east`). Same question for other direction cases

java/bogdana_maksiuta/src/com/zoolatech/lecture2/tasks/_2/Robot.java
```diff
@@ -0,0 +1,189 @@
...
+                } else {
+                    int allowed = getWidthRoom() - getW() - 1;
```
tromanyuk, line 79: I think you should change the coordinates by a number of allowed moves

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Musician.java
```diff
@@ -0,0 +1,40 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.Objects;
+
+public class Musician {
+    private String name;
+    MusicalInstrument instrument;
```
tromanyuk, line 7: `instrument` should be defined as private

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_1/Musician.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public Musician(String name, MusicalInstrument instrument) {
+        this.name = name;
+        this.instrument = instrument;
+    }
+
+    public void getNameAndInstrument(Musician musician) {
```
tromanyuk, line 14:
```
A musician should also have a method which should print his/her name and the name of an instrument.
```

The idea was to print a current musician. In other words, it was enough to provide a method `toString`

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/NumberInTheRangeValidation.java
```diff
@@ -0,0 +1,16 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public class NumberInTheRangeValidation implements Validator {
+    private int birthMonth;
```
tromanyuk, line 4: The class name is `NumberInTheRangeValidation` and the instance field name - `birthMonth`. The idea was the instance field will be named like `fieldName` and the class will be called like this:

```
new NumberInTheRangeValidation("birthMonth"), new NumberInTheRangeValidation("birthDay")...
```

you could even add possible ranges as additional arguments of the constructor

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/StringIsNotEmptyValidation.java
```diff
@@ -0,0 +1,25 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public class StringIsNotEmptyValidation implements Validator {
+    private String firstName;
```
tromanyuk, line 4: this validator should be used only for a single field at once

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccount.java
```diff
@@ -0,0 +1,81 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import java.util.Objects;
+
+public class UserAccount {
```
tromanyuk, line 5: this class could be defined as record as well

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,33 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import java.util.ArrayList;
+
+public class UserAccountValidator implements Validator {
+    UserAccount userAccount;
+
+    public UserAccountValidator(UserAccount userAccount) {
```
tromanyuk, line 8: The validator should accept a user account as a method argument

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
```diff
@@ -0,0 +1,33 @@
...
+        this.userAccount = userAccount;
+    }
+
+    @Override
+    public boolean isValid(UserAccount userAccount) {
+        ArrayList<Validator> userAccountValidators = new ArrayList<>();
```
tromanyuk, line 14: validators can be added to the list in the constructor (note, as per task description you need to pass user account field names, and not values).

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_2/UserAccountValidator.java
tromanyuk: 1. we need to validate id, but in a different manner 
2. usually, it's better to define them in separate files (it's simpler to read code if file doesn't have a lot of it)

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_3/Circle.java
```diff
@@ -0,0 +1,44 @@
+package com.zoolatech.lecture3.tasks._3;
+
+import java.util.Objects;
+
+public non-sealed class Circle implements Shape {
+    final static double PI = 3.1415926536;
```
tromanyuk, line 6: private

java/bogdana_maksiuta/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
```diff
@@ -0,0 +1,34 @@
...
+    String fileData;
+    String fileName;
+
+    public Reporter(String headerData, String fileData, String fileName) {
+        this.headerData = headerData;
+        this.fileData = fileData;
+        this.fileName = fileName;
+    }
+
+    void generate() {
```
tromanyuk, line 14: Almost correct:
1. Children classes don't need to override method generate - they should used the version of a parent class
2. Methods `addHeaderData` and `addFileData` should be abstract in the parent class, and overridden in children classes

tromanyuk:
to remove .class files, you can add folder "out" to gitignore, remove files from your local machine, then commit and push changes

PR #101, opened 2022-07-23
Anton Kozyk HW-3

AntonKun:


java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_1/MusicalInstruments.java
```diff
@@ -0,0 +1,40 @@
+package com.zoolatech.lecture3.tasks._1;
+
+abstract public class MusicalInstruments {
```
tromanyuk, line 3: 1. usually, the access modifier is defined first: `public abstract`
2. prefer to use a singular form for class names (MusicalInstrument, without s)

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_1/MusicalInstruments.java
```diff
@@ -0,0 +1,40 @@
...
+    String instrumentName;
+    abstract void playSound();
+}
+
+class Guitar extends MusicalInstruments {
+    Guitar() {
+        instrumentName = "Guitar";
```
tromanyuk, line 10: it would be better to use a constructor of a parent class (since the variable `instrumentName` is defined there)

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_1/Musician.java
```diff
@@ -0,0 +1,85 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.Objects;
+
+abstract public class Musician {
+    String name;
```
tromanyuk, line 6: `private` - don't forget regarding the encapsulation

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_1/Musician.java
```diff
@@ -0,0 +1,85 @@
...
+        Musician musician = (Musician) o;
+        return Objects.equals(name, musician.name);
+    }
+
+    void printInfo() {
+        System.out.println("Name: " + name);
+    }
+}
+
+class Guitarist extends Musician {
```
tromanyuk, line 27: From the code point of view, what is the difference between classes `Guitarist`, `Pianist`, and `Violinist`? Do we really need to have separate classes for them? Just for comparison: classes that extend `MusicalInstruments` have different implementations of the method `playSound`, so it's fine to keep the hierarchy there.

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_1/Orchestra.java
```diff
@@ -0,0 +1,23 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.ArrayList;
+
+public class Orchestra {
+    private ArrayList<Musician> orchestra = new ArrayList<>();
+
+    void add(Musician musician) {
+        orchestra.add(musician);
+        System.out.println(musician.name + " added to orchestra");
```
tromanyuk, line 10:
```
musician.name
```

prefer to not access other class variables directly

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_2/MainValidator.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+public class MainValidator implements Validating {
+    private List<Validator> validators = new ArrayList<>();
+    private boolean validationResult = true;
+
+    @Override
+    public boolean isValid(UserAccount account) {
+        validators.add(new StringIsNotEmptyValidator("First name"));
```
tromanyuk, line 12: It would be better to add these validators during the object construction. Right now, new validator instances will be added to the list on each call of the `isValid` method

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_2/MainValidator.java
```diff
@@ -0,0 +1,31 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class MainValidator implements Validating {
+    private List<Validator> validators = new ArrayList<>();
+    private boolean validationResult = true;
```
tromanyuk, line 8: I don't think that `validationResult` should be considered as a part of the MainValidator state. I guess it would be better to define it as a local variable.

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_2/MainValidator.java
```diff
@@ -0,0 +1,31 @@
...
+        validators.add(new StringIsNotEmptyValidator("Country"));
+        validators.add(new EmailValidator("Email"));
+        validators.add(new PhoneNumberValidator("Phone number"));
+        validators.add(new NumberIsInRangeValidator("Birth day"));
+        validators.add(new NumberIsInRangeValidator("Birth month"));
+        validators.add(new NumberIsBiggerValidator("User ID"));
+
+        for (Validator validator : validators) {
+            if (!validator.isValid(account)) {
+                validationResult = false;
```
tromanyuk, line 23: note: the result can be returned directly from this line of code:

```
return false;
```

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_2/Validating.java
```diff
@@ -0,0 +1,5 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public interface Validating {
```
tromanyuk, line 3: it's better to use nouns or adjectives for interface names (e.g., Validator)

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_3/Figure.java
```diff
@@ -0,0 +1,9 @@
+package com.zoolatech.lecture3.tasks._3;
+
+public abstract class Figure {
```
tromanyuk, line 3: What is the reason to provide this class? As for me, it looks the same as the interface Shape.

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_3/Triangle.java
```diff
@@ -0,0 +1,63 @@
...
+
+    Triangle(double sideA, double sideB, double sideC){
+        this.sideA = sideA;
+        this.sideB = sideB;
+        this.sideC = sideC;
+    }
+
+    @Override
+    public double findPerimeter() {
+        double trianglePerimeter = sideA + sideB + sideC;
```
tromanyuk, line 18: since it's clear from the method name what will be the result, you can return the value directly here, without introducing of a local variable

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_3/Figure.java
tromanyuk, line 3: I think this class can be defined as an interface

java/Anton-Kozyk/src/com/zoolatech/lecture3/tasks/_4/Reporter.java
```diff
@@ -0,0 +1,36 @@
...
+    String fileData;
+    String fileName;
+
+    Reporter(String headerData, String fileData, String fileName) {
+        this.headerData = headerData;
+        this.fileData = fileData;
+        this.fileName = fileName;
+    }
+
+    void generate() {
```
tromanyuk, line 14: probably, this method same as constructor can be public

tromanyuk (approved):

PR #103, opened 2022-07-25
Cherepnin Oleh: homework for lecture3

Oleghon:


java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_1/Drum.java
```diff
@@ -0,0 +1,25 @@
...
+
+public class Drum extends MusicInstrument {
+
+    public Drum(String name) {
+        super(name);
+    }
+
+    @Override
+    public String makeSound() {
+        return new StringBuilder(getName()).append(": bum-bum").toString();
```
tromanyuk, line 11: you can just use the concatenation operator here

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_1/Drum.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "Drum{ " + getName() + " }";
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
tromanyuk, line 20: what about hashCode? 😄

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_1/MusicInstrument.java
```diff
@@ -0,0 +1,30 @@
...
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof MusicInstrument)) return false;
+        MusicInstrument that = (MusicInstrument) o;
+        return getName().equals(that.getName());
```
tromanyuk, line 23: you can access variable `name` directly here; same for `hashCode`

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_1/Orchestra.java
```diff
@@ -0,0 +1,28 @@
+package com.zoolatech.lecture3.tasks._1;
+
+import java.util.Set;
+
+public class Orchestra {
+    private Set<Musician> musicians;
+
+    public Orchestra(Set<Musician> musicians) {
```
tromanyuk, line 8: probably, it might be better to create a set instance inside of this class instead of passing it in the constructor to not expose its state

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/DateValidator.java
```diff
@@ -0,0 +1,33 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public class DateValidator implements SmallerValidator {
```
tromanyuk, line 3: The smaller validators need to accept **`a field name`** during the creation

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NameValidator.java
```diff
@@ -0,0 +1,30 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import java.lang.reflect.Field;
+
+public class NameValidator implements SmallerValidator {
```
tromanyuk, line 5: You needed to define a class for each validation rule:

```
List of validation rules: string is not empty, valid email, valid phone number, number is in the range between X and Y, number is bigger than a value X, number is less than a value X. 
```

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NameValidator.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private String regex = "[A-Z][a-z]+";
+
+    @Override
+    public boolean isValid(User user) {
+        try {
+            boolean firstNameCheck = printMessage(user.getClass().getDeclaredField("firstName"), user.getFirstName());
```
tromanyuk, line 12:
```
user.getClass().getDeclaredField("firstName")
```

You've already defined a field name as a string literal. Why do you need to retrieve this field through reflection and get the same value?

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NameValidator.java
```diff
@@ -0,0 +1,30 @@
...
+        } catch (NoSuchFieldException e) {
+            e.printStackTrace();
+            return false;
+        }
+    }
+
+    private boolean printMessage(Field field, String name) {
+        if (name.isBlank()) {
+            System.out.println(field.getName() + " is blank");
+        } else if (!name.matches(regex)) {
```
tromanyuk, line 25: It's better to create an instance of the class Pattern and reuse it instead of using `String.matches`. Method `matches` will create a new instance of the Pattern during every invocation, and this is a heavy operation.

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/UserValidator.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.List;
+
+public class UserValidator {
+    private static final List<SmallerValidator> VALIDATORS = List.of(
+            new NameValidator(),
+            new NumberValidator(),
+            new DateValidator(),
+            new EmailValidator());
+
+    public static boolean validate(User user) {
```
tromanyuk, line 12:
```
All validators need to have a method isValid that accepts a user account object and returns either true or false.
```

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/UserValidator.java
```diff
@@ -0,0 +1,24 @@
+package com.zoolatech.lecture3.tasks._2;
+
+import java.util.List;
+
+public class UserValidator {
+    private static final List<SmallerValidator> VALIDATORS = List.of(
```
tromanyuk, line 6: Why static?

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NumberValidator.java
```diff
@@ -0,0 +1,17 @@
+package com.zoolatech.lecture3.tasks._2;
+
+public class NumberValidator implements SmallerValidator {
+
+    private String regex = "\\d{3}-\\d{3}-\\d{2}-\\d{2}";
```
tromanyuk, line 5: can be constant

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_3/Triangle.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public double findPerimeter() {
+        // P = a + b + c
+        if (checkTriangle()) {
+            return aLength + bLength + cLength;
+        }
+        throw new ArithmeticException("Triangle does not exist");
+    }
+
+    private boolean checkTriangle() {
```
tromanyuk, line 30: Probably, this method should be called during the object construction. Also, it's better to use `IllegalArgumentException` instead of `ArithmeticException`

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_4/AbstractReporter.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    public void generate() {
+        System.out.println("Opening a file: " + name +
+                "\nAdd header:" + header +
+                "\nAdd data:" + data);
+        convert();
+    }
+
+    protected abstract void convert();
```
tromanyuk, line 25: Why do we need to have this method? As per task description:

```
and have a method called generate. The method needs to create a new file, add header data in an appropriate format (CSV, PDF, DOC), add file data in an appropriate format and close a file. ...  like “Opening a file test.txt” or “Adding file details in CSV format” for each action....
```

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/DateValidator.java
Oleghon: Why is this field needed?

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NameValidator.java
Oleghon: > number is in the range between X and Y, number is bigger than a value X, number is less than a value X.
Should I split DateValidator class into several other?

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/DateValidator.java
tromanyuk: The field is simply used for the output like
```
“validating that fieldName is less than 10”
```

java/Cherepnin_Oleh/src/com/zoolatech/lecture3/tasks/_2/NameValidator.java
tromanyuk: Yes. You don't need to implement a validator for a specific field/fields. You need to implement validators for specific validation rules.

tromanyuk (approved):

PR #122, opened 2022-08-04
Cherepnin Oleh: homework for lecture 7

Oleghon:


java/Cherepnin_Oleh/src/com/zoolatech/lecture7/tasks/_2/SyncCalculator.java
```diff
@@ -0,0 +1,25 @@
+package com.zoolatech.lecture7.tasks._2;
+
+public class SyncCalculator implements Calculator {
+    private volatile int value;
```
tromanyuk, line 4: you don't need to define a variable as `volatile` if all operation that work with it are protected with the synchronized block - it provides a same guarantees (memory values will be synchronized)

java/Cherepnin_Oleh/src/com/zoolatech/lecture7/tasks/_2/SyncCalculator.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public void add(int addedValue) {
+        synchronized (this) {
```
tromanyuk, line 7: it would be simpler to define the whole method as synchronized (it will be the same, but you'll need to write less code)

tromanyuk (approved):

PR #126, opened 2022-08-05
Homework JS #1 by Anastasiia Afanasieva

anastasiiaafanasieva:
Implement game logic and improve styles for it


zoolamyke:
# Review #126:

- It's a clever way of deriving the item quantity, but it's much less reliable than `meshContainer.children.length`
image

- It's okay to combine operations, but it's almost always better to make operations separately
image

E.g.
```
const currentScore -= clickOutsideElementPoint
scoreContainer.innerText = currentScore
```

- It would be more comfortable to have `upperLimit` and `lowerLimit`
image

- Just in case, remember to also remove event listeners
image

- This chunk of logic is duplicated
image

- To be more direct, convert to number via `Number(value)`
image

- You don't need `meshSize` here - `return variants[getRandomNumber(variants.length - 1)];` should be good enough.
image

- This `return` is not used
image

- Can we use `deactivateElement` here?
image

- `Boolean(!something)` is converting boolean to boolean. `!something` is sufficient.
image

zoolamyke:
Чудовий дизайн, хороша робота.

zoolamyke:
https://user-images.githubusercontent.com/54806213/184590992-ae0b1740-b89b-495b-b605-8d76c123b8b3.mov

anastasiiaafanasieva:
@zoolamyke Hey! I resolved comments and added `.gitignore`.

PR #127, opened 2022-08-05
Oleksandr Nikiforov Hw7

olnikiforov:


java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/Calculator.java
```diff
@@ -0,0 +1,21 @@
+package com.zoolatech.lecture7.tasks._2;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public abstract class Calculator {
```
tromanyuk, line 5: Do you really need to have this base class? You're overriding it's methods in all three subclasses, and for the atomic version the value even stored in a different variable.

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/Calculator.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public void add(int valueToAdd) {
+        value += valueToAdd;
+    }
+
+    public void subtract(int valueToSubtract) {
+        value -= valueToSubtract;
+    }
+
+    public abstract <T> T getValue();
```
tromanyuk, line 20: other methods of this class work with int values, and this method is generic - you need to make all methods to work with int values, or make all methods generic

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/AtomicCalculator.java
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    @Override
+    public void subtract(int valueToSubtract) {
+        atomicValue.addAndGet(-valueToSubtract);
+        System.out.println("Subtract - " + valueToSubtract);
+    }
+
+    @Override
+    public AtomicInteger getValue() {
```
tromanyuk, line 26: you need to return an actual value (int)

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/SynchronizedCalculator.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class SynchronizedCalculator extends Calculator {
+
+    public SynchronizedCalculator(int value) {
+        super(value);
+    }
+
+    @Override
+    public void add(int valueToAdd) {
+        synchronized (this) {
```
tromanyuk, line 13: it's better to define the whole method as synchronized - the functionality remains the same, but it requires less code

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/AtomicCalculator.java
```diff
@@ -2,29 +2,25 @@
...
     public AtomicCalculator(int value) {
         super(value);
-        this.atomicValue = new AtomicInteger(value);
     }
 
     @Override
     public void add(int valueToAdd) {
+        AtomicInteger atomicValue = new AtomicInteger(value);
         atomicValue.addAndGet(valueToAdd);
+        value = atomicValue.get();
```
tromanyuk, line 15: this doesn't prevent race conditions: two might enter this method in the same moment of time, and perform this operation concurrently

suggestion: remove this class from hierarchy and simply store the value in the AtomicInteger variable

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/ExplicitLocksCalculator.java
```diff
@@ -2,36 +2,30 @@
 
 import java.util.concurrent.locks.ReentrantLock;
 
-public class ExplicitLocksCalculator extends Calculator{
+public class ExplicitLocksCalculator extends Calculator {
```
tromanyuk, line 5: getter method should use locks as well

java/Nikiforov-Oleksandr/src/com/zoolatech/lecture7/tasks/_2/SynchronizedCalculator.java
```diff
@@ -0,0 +1,20 @@
+package com.zoolatech.lecture7.tasks._2;
+
+import java.sql.SQLOutput;
+
+public class SynchronizedCalculator extends Calculator {
```
tromanyuk, line 5: getter method (`getValue`) should be synchronized as well

tromanyuk (approved):

PR #132, opened 2022-08-07
Homework JS #1 by Maksym Omelianenko

Omax0:
Completed game with main rules and added new features for the more interest.


zoolamyke:
# Review #132:

- Use `background-color: transparent`
image

- Do not use `alert`, `prompt`, `confirm` - they're all way too outdated in terms of UX, and they stop JS execution which is not great either

- For some reason, I didn't find `package.json` so I had to copy it from a different project

- Prefer `class` over `style`
image

- This function appears to not be used anywhere
image

- In non-hardcore mode, there is a bug: after the element is clicked, the round duration is reset and the round is effectively longer than if I didn't click on the element

zoolamyke:
Хороша робота

zoolamyke:
https://user-images.githubusercontent.com/54806213/184591088-227f1184-0a8a-4bb5-b921-b1d51744d70c.mov

PR #133, opened 2022-08-07
[JS-HW1] Gamid Akhmedov

Gamid-FrontEnd:


zoolamyke:
# Review #133:

- Avoid using raw numbers - use constants instead
image

- `onClick` -> `onGameStart`
image

- It's a clever way of detecting an active element, but it's not reliable as CSS classes can be changed by someone who doesn't know they are being used as identifiers. But you can use `data-` attributes for such a purpose.
image

- Avoid using raw numbers - use constants instead
image

- This chunk of code is duplicated multiple times
image

- Prefer `class` over `style`
image

- `meshContainer` event listeners are a bit hard to follow. If you've chosen a one-listener approach, you can just set a listener **once** and only remove it when the game is over.
image

zoolamyke:
Гарний дизайн, хороша робота. Є кілька багів, один серйозний - табло оновлюється лише наприкінці раунду при міскліках, а повинно одразу.

zoolamyke:
https://user-images.githubusercontent.com/54806213/184591184-e6974aa3-acbc-465c-86a5-99f621479a32.mov

PR #134, opened 2022-08-08
HW8 Java Efe Adzhar

EfeAdzhar:


.gitignore
```diff
@@ -1,4 +1,9 @@
 .idea/
 **/.DS_Store
+/java/Anton-Kozyk/Anton-Kozyk.iml
```
tromanyuk, line 3: 1. Please remove `/java/Anton-Kozyk/Anton-Kozyk.iml` from the file
2. `/out/` is defined twice

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture8/tasks/_1/Number1.java
```diff
@@ -0,0 +1,124 @@
...
+(the process can simply print an appropriate message and sleep for a couple of seconds).
+When an order is ready, it should be passed to the first available DeliveryPerson, which needs to deliver an order to a customer
+(also, just print an appropriate message and sleep for some time).
+Use queues to synchronize order processing between customers, chefs and delivery persons.
+Note: both chefs and delivery persons need to run in the separate thread.
+*/
+
+public class Number1 {
+    public static void main(String[] args) throws InterruptedException {
+        Lock lock = new ReentrantLock();
```
tromanyuk, line 22: `lock` is unused and can be removed

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture8/tasks/_1/Number1.java
```diff
@@ -0,0 +1,124 @@
...
+            return order.getId() + " " + order.getCustomerName();
+        } finally {
+            lock.unlock();
+        }
+    }
+}
+
+class Order {
+    AtomicInteger id;
+    volatile String customerName;
```
tromanyuk, line 100: 1. `customerName` isn't modified from other threads. So, we don't need to define it as `volatile`. 
2. `customerName` doesn't change after object creation, so it can be defined as final
3. all instance variables can be defined as private
4. id can be just an `int` field - it doesn't change after object creation (same it should be private and final)
5. since getOrder works with immutable data - explicit lock can be removed

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture8/tasks/_1/Number1.java
```diff
@@ -0,0 +1,124 @@
...
+
+        customerQueue.start();
+        chef.start();
+        deliveryPerson.start();
+
+    }
+}
+
+class Restaurant {
+    Lock lock = new ReentrantLock();
```
tromanyuk, line 86: same: order has only immutable data - lock can be removed

tromanyuk (approved):

.gitignore
```diff
@@ -1,8 +1,16 @@
 .idea/
 **/.DS_Store
+<<<<<<< HEAD
```
tromanyuk, line 3:
```
<<<<<<< HEAD
=======
>>>>>>> master
```
these lines should be removed from the file; also, there are duplicated definitions in the file

tromanyuk (approved):

PR #139, opened 2022-08-09
Move files to my folder

sonyachuyeva:


zoolamyke:
# Review #139:

- Always create new code over existing, don't duplicate it elsewhere! It makes it hard to review, as GitHub won't show the diff correctly
- `itemsAmount` and `meshSize` duplicate each other, there should be only one source of truth.
image
- `getIndex` -> `getNextActiveIndex`
- This chunk is duplicated
image
- Use verbs, not nouns, when naming functions
image
- This chunk is duplicated multiple times
image
- The approach with `e.detail` is tricky, but will only work for double clicks, so it's not reliable
image
- `element.removeEventListener` should be called inside `deactivateElement`, makes it more maintainable.
image
- Please, don't use `alert`
- Prefer `class` over `style`
image
- `meshContainer` -> `document`. The idea is to prevent any clicks outside target element.
image
- Don't use `==` ever
image
- It makes sense to move `clearInterval` inside `immediateInterval` (which should be renamed to something like `runImmediate`)
image

zoolamyke:
Хороша робота! Знайшов баги якщо трохи попрацювати з різними швидкостями.

zoolamyke:
https://user-images.githubusercontent.com/54806213/184591440-0957cfeb-dc74-4167-9c1b-0017b3b91cb6.mov

PR #140, opened 2022-08-09
[JS-HW1] Nikita Kucherenko

nikson666:


zoolamyke:
# Review #140:

- Only add a description if there is something non-obvious
image

- Use `classList` over `style` attribute
image

- Array of children can be accessed via `meshContainer.children`. Also, use `querySelector` and `querySelectorAll` when searching for elements. Other methods are outdated.
image

- Remove `console.log`
image

- Avoid using raw numbers - use constants instead
image

- Use `for/of` or `arr.forEach`
image

- Avoid detecting what is what based on a classname - it is not reliable.
image

- Ternary operator cannot be used in such manner (it is not a replacement for `if/else`)
image

Here's how it should work:
```
        scoreContainer.style.background = currentScore < 0 ?  "red" : "green"
```

- `onActiveElement` -> `onActiveElementClick`. The same goes for  `onOutsideActiveElement`
image

zoolamyke:
https://user-images.githubusercontent.com/54806213/184591513-6cce8753-dca9-4128-8c79-f3e510886936.mov

PR #142, opened 2022-08-09
[JS-HW1] Kharin Leonid

leonid-06:


zoolamyke:
# Review #142:

- Use Prettier in your IDE

- Avoid creating additional variables in `for` loop header; also, avoid using raw numbers - use constants instead
```
const meshSize = 25
for (let i = 0; i < meshSize; i++) {
  const el = document.createElement('div')
  ...
}
```
image
image


image

- Use verbs, not nouns, when naming functions! E.g. `setCurrentScoreColor`
image
image
image

- `meshContainer` listener can be set just once, there's no benefit to adding/removing it every second
image


- Name event handlers correctly - ideally, with `on` prefix - `onSomethingClick`, or `onSomethingChange`, or `handleSomeChange`
image

- Always create new code over existing, don't duplicate it elsewhere! It makes it hard to review, as GitHub won't show the diff correctly

- Never use alerts

- Use `let` to define variables
image

- Similar behavior can be achieved without an array, try to figure it out.
image

- Pay more attention to naming
image

- Remove `console.log` entries
image

zoolamyke:
Хороший редизайн! Немає багів. Хороша робота.

zoolamyke:
https://user-images.githubusercontent.com/54806213/184591921-de4ffac6-6219-4b65-85bf-2dd207c4342c.mov

PR #146, opened 2022-08-11
Anton Kozyk HW10

AntonKun:


java/Anton-Kozyk/src/com/zoolatech/lecture10/tasks/_1/FileReader.java
```diff
@@ -0,0 +1,39 @@
...
+                System.out.println("-" + file);
+            }
+        }
+    }
+
+    public void fileContent(File file) throws IOException {
+        if (!file.exists()) {
+            System.err.println("File '" + file.getName() + "' not found. Try again...\n");
+            try (FileWriter fileWriter = new FileWriter(directory.getPath() + "/error.log", true)) {
+                fileWriter.write(file.getName() + " [" + new Date() + "]\n");
```
ogorbov, line 30: Please use java 8 date-time classes. util.Date is an old approach. More context here https://programminghints.com/2017/05/still-using-java-util-date-dont/

ogorbov (commented):
Just general feedback. Overall the implementation is working. But as you tried to extract some code to a separate class here are a few notes:
- FileReader is not a good name for a class because it can write something to a log error file. Reader starts to write. 
- directoryContent() and fileContent() also not clear. You need to open them to understand what they are doing. I feel that verb like print can fix that.
- Interesting idea with do/while. I cannot say that you can see it in code frequently, but it looks good here!

java/Anton-Kozyk/src/com/zoolatech/lecture10/tasks/_1/FileReader.java
AntonKun: Done, now it uses LocalDateTime.
Thank you, that's very useful information for me! And thanks for the link.

ogorbov (approved):

PR #148, opened 2022-08-12
HW-10-I/O-Java-Efe-Adzhar

EfeAdzhar:


java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,54 @@
...
+            default:
+                //creating error.log file and writing down incorrect scanner input(file name)
+                errorLog(file, string);
+        }
+    }
+
+    private static void errorLog(File file, String string) throws IOException, NoFileException {
+        int counter = 0;
+        try (FileWriter fileWriter = new FileWriter(file + "/error.log",true)){
+            fileWriter.write(string + " invalid input number: "+ new Date() + "\n");
```
ogorbov, line 49: Please use dates from java 8. https://programminghints.com/2017/05/still-using-java-util-date-dont/ util.Date is depricated

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,54 @@
...
+                System.out.println("File number " + ++counter + " : " + value.getName());
+            } else if (value.isDirectory()) {
+                System.out.println("Directory: " + value.getName());
+            } else {
+                throw new NoFileException();
+            }
+        }
+    }
+
+    public static void generateFiles(String string, File file) throws IOException, NoFileException {
```
ogorbov, line 22: Please use some meaningful name instead of string and file

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,54 @@
...
+        switch (string) {
```
ogorbov, line 23: Let's refactor this switch to be more generic. Current implementation depends on the file names. You can check if listFiles contains name provided by user and if not just write error.

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/NoFileException.java
```diff
@@ -0,0 +1,7 @@
+package com.zoolatech.lecture10.tasks._1;
+
+final class NoFileException extends Error {
```
ogorbov, line 3: 2 thoughts on that:
Error is a very specific type of erorrs. Let's say that we are not talking about checked exceptions.  You have option to define custom exception which is unchecked (RuntimeException) or to use Error. The main difference is that if a client of this code might want to catch it - use unchecked one. If client cannot fix this problem - use Error (extremly rarely used). Common example of errors are OutOfMemoryError (it means that client cannot fix the problem. you don't have memory and you cannot allocate it in runtime). So in your case extends RuntimeException would be better. But I'm strongly belive that it's better to avoid custom exception if you have one in jdk and don't need special handling. I'd use FileNotFoundException which is part of jdk.

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,54 @@
...
+                break;
+            default:
+                //creating error.log file and writing down incorrect scanner input(file name)
+                errorLog(file, string);
+        }
+    }
+
+    private static void errorLog(File file, String string) throws IOException, NoFileException {
+        int counter = 0;
+        try (FileWriter fileWriter = new FileWriter(file + "/error.log",true)){
```
ogorbov, line 48: nit: I think you need to run formatting. Command+Option+L if you use mac

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/db/surnames.txt
```diff
@@ -0,0 +1,3 @@
+Adzhar
+Fung
+Scott
```
ogorbov, line 3: Please remove .class files

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
...
+    private boolean errorLog(String fileName) throws IOException {
+        final LocalDate localDate = LocalDate.now();
+        final LocalDateTime localDateTime = LocalDateTime.now();
+        try (FileWriter fileWriter = new FileWriter(file + "/error.log", true)) {
+            fileWriter.write(fileName + " invalid input was made: " + localDate + ", time: " + localDateTime + "\n");
+        }
+        System.out.println("No file with such name");
+        System.out.println("Written invalid filename to error.log");
+        System.out.println("1. Start\n2. Exit");
+        return false;
```
ogorbov, line 51: Very tricky logic here. Let's make this method void and return false in generateFiles. Because You are trying to say that generateFiles was not successful  but errorLog was good

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
...
+                    FileInputStream fileInputStream = new FileInputStream(file + "/" + fileName);
+                    String result = new String(fileInputStream.readAllBytes());
+                    System.out.println(result);
+                    return true;
+                }
+            }
+            return errorLog(fileName);
+    }
+
+    private boolean errorLog(String fileName) throws IOException {
```
ogorbov, line 42: Please use verb at the beginning for any method. logError will be better

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
...
+        final LocalDate localDate = LocalDate.now();
+        final LocalDateTime localDateTime = LocalDateTime.now();
+        try (FileWriter fileWriter = new FileWriter(file + "/error.log", true)) {
+            fileWriter.write(fileName + " invalid input was made: " + localDate + ", time: " + localDateTime + "\n");
+        }
+        System.out.println("No file with such name");
```
ogorbov, line 48: Please use Sustem.err here

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
ogorbov: However, I don't think that you need to return a boolean. You are already notifying user with system.out

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
...
+            }
+        }
+    }
+
+    public boolean generateFiles(String fileName) throws IOException {
+        File[] files = file.listFiles();
+        assert files != null;
+            for (File fileInFiles : files) {
+                if (fileInFiles.exists() && fileInFiles.getName().equals(fileName)) {
+                    FileInputStream fileInputStream = new FileInputStream(file + "/" + fileName);
```
ogorbov, line 33: FileInputStream is not closed

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
...
+        for (File value : files) {
+            if (value.isFile()) {
+                System.out.println("File number " + ++counter + " : " + value.getName());
+            }
+        }
+    }
+
+    public boolean generateFiles(String fileName) throws IOException {
+        File[] files = file.listFiles();
+        assert files != null;
```
ogorbov, line 30: it's better to print some error message to System.err stream instead of assert. To be honest it's vary rarely used in production code

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,53 @@
+package com.zoolatech.lecture10.tasks._1;
+
+import java.io.*;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.Arrays;
```
ogorbov, line 6: Please remove import

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/Menu.java
```diff
@@ -0,0 +1,16 @@
+package com.zoolatech.lecture10.tasks._1;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Scanner;
+
+final class Menu {
+    public static void menu() throws IOException {
```
ogorbov, line 8: Please do not use static here. just new Menu().menu(). It's not a mistake but static methods are always a problem for testing.

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,61 @@
...
+        File[] files = file.listFiles();
+        if (files == null) {
+            System.err.println("File is empty");
+        } else {
+            for (File fileInFiles : files) {
+                if (fileInFiles.exists() && fileInFiles.getName().equals(fileName)) {
+                    try (FileInputStream fileInputStream = new FileInputStream(file + "/" + fileName)) {
+                        String result = new String(fileInputStream.readAllBytes());
+                        System.out.println(result);
+                        fileInputStream.close();
```
ogorbov, line 40: Please revise this article. https://www.baeldung.com/java-try-with-resources

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,61 @@
...
+                        System.out.println("1. Start\n2. Exit");
+                    }
+                } else {
+                    logError(fileName);
```
ogorbov, line 44: Let's say you have 3 files. If you select the third one it means that before the file will be found the user will see a few error messages and a few times error will be logged. This part can be optimized and we need to log error only if listFiles doesn't have the desired file.

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public FileRedactor(File file) {
+        this.file = file;
+    }
+
+    public void showFiles() {
+        File[] files = file.listFiles();
+        int counter = 0;
+
+        assert files != null;
```
ogorbov, line 22: Please use System.err as in next method

java/Mekhmet-Efe_Adzhar/src/com/zoolatech/lecture10/tasks/_1/FileRedactor.java
```diff
@@ -0,0 +1,61 @@
...
+
+final class FileRedactor {
+
+    private final File file;
+
+    public FileRedactor(File file) {
+        this.file = file;
+    }
+
+    public void showFiles() {
```
ogorbov, line 18: nit: 
showFiles  -> showAllAvailableFiles
generateFiles -> showFileContent

ogorbov (approved):

PR #150, opened 2022-08-14
Java Homework#4 Solve tasks for HW4

stymoshchuk:


java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_1/Container.java
```diff
@@ -0,0 +1,40 @@
+package com.zoolatech.lecture4.tasks._1;
+
+import java.util.Objects;
+
+public class Container {
+    String anyValue;
```
tromanyuk, line 6: instance field should be defined as `private`

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_1/Container.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public String getAnyValue() {
+        return anyValue;
+    }
+
+    public void setAnyValue(String anyValue) {
+        this.anyValue = anyValue;
+    }
+
+    public Container(String anyValue){
```
tromanyuk, line 16: constructors should be defined before methods

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_1/Container.java
```diff
@@ -0,0 +1,40 @@
...
+        this.anyValue = anyValue;
+    }
+
+    class ContainerUpdatesListener{
+        String newAnyValue;
```
tromanyuk, line 21: same: private

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_1/Container.java
```diff
@@ -0,0 +1,40 @@
...
+
+        public ContainerUpdatesListener(){
+            Container.this.anyValue = newAnyValue;
```
tromanyuk, line 24: this should be opposite: newAnyValue = anyValue;

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_1/Container.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public boolean hasValueChanged(){
+            return !Objects.equals(getAnyValue(), newAnyValue);
```
tromanyuk, line 28: you can access the field directly, without a method

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_3/StringModifier.java
```diff
@@ -0,0 +1,6 @@
+package com.zoolatech.lecture4.tasks._3;
+
+@FunctionalInterface
+interface StringModifier {
+    abstract String modify(String initialString);
```
tromanyuk, line 5: all interface methods are `abstract` by default, you don't need to add this modifier

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_3/Main.java
```diff
@@ -0,0 +1,14 @@
+package com.zoolatech.lecture4.tasks._3;
```
tromanyuk, line 1: second task is missing

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_3/Main.java
```diff
@@ -0,0 +1,14 @@
...
+
+public class Main {
+    public static void main(String[] args) {
+        test("String");
+    }
+        private static void test(String test){
```
tromanyuk, line 7:
```
 In the main class create a method that accepts a string value and a lambda expression, applies the lambda to the value and outputs the result. 
```

method should accept lambda as well

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_3/Main.java
```diff
@@ -0,0 +1,14 @@
...
+            StringModifier stringModifier = initialString -> initialString.toUpperCase();
```
tromanyuk, line 8:
```
Both lambdas should prepend a prefix “modified: ” before the modified values
```

modified should be the part of a lambda definition

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_3/Main.java
```diff
@@ -0,0 +1,14 @@
...
+            StringModifier stringModifier1 = initialString -> initialString.toLowerCase();
+            System.out.printf("Modiifed: %s", stringModifier.modify(test)+"\n");
```
tromanyuk, line 10:
```
Call the same method with two method references that just convert all characters to the lower and upper case.
```

this part wasn't completed

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_4/Unit.java
```diff
@@ -0,0 +1,93 @@
...
+        this.aliases = aliases;
+    }
+
+    public String getAbbreviation() {
+        return abbreviation;
+    }
+
+    public static Unit getByAlias(String alias) {
+        for (Unit unit : Unit.values()) {
+            if (unit.hasAlias(alias.toLowerCase())) return unit;
```
tromanyuk, line 28:
```
unit.name().equals(alias.toUpperCase())
```
OR
```
unit.name().equalsIgnoreCase(alias)
```
would allow you to get rid of the class `Aliases`

tromanyuk, line 28: please add braces for single line if conditions

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_4/Unit.java
```diff
@@ -0,0 +1,93 @@
...
+            case POUND -> result = value * 453.6;
+            case TON -> result = value * 907185;
+            case GRAM -> result = value;
+        }
+        return result;
+
+    }
+
+    public static class Aliases {
+        String[] aliases;
```
tromanyuk, line 69: you could also define `aliases` as a list instead of an array - that would allow you to get rid of most of methods, and even define just list of values directly inside of the Unit class without creation of an additional class

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_4/Unit.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+        return null;
+    }
+
+    public boolean hasAlias(String al) {
+        if (aliases == null) return false;
+        else return aliases.contains(al);
+    }
+
+    static double convertValue(double value, Unit necessaryValue, Unit initialUnit) {
```
tromanyuk, line 38: 1. the method shouldn't be static
2. the method should accept two arguments: value and initialUnit (source unit)
3. necessaryValue is current object (this)

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_4/Unit.java
```diff
@@ -0,0 +1,93 @@
...
+        double convertedValue = convertValueToG(initialUnit, value);
+        return convertValueToDesiredUnit(necessaryValue, convertedValue);
+    }
+
+    static double convertValueToDesiredUnit(Unit necessaryValue, double value) {
+        double result = 0;
+        switch (necessaryValue) {
```
tromanyuk, line 45:
```
return switch(necessaryValue){
    case KILOGRAM -> value / 1000;
    case GRAM ->  value;
    ...
};
```

same for the method `convertValueToG`

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_7/StringRepresentation.java
```diff
@@ -0,0 +1,44 @@
...
+        numbers.put('7', "(seven)");
+        numbers.put('8', "(eight)");
+        numbers.put('9', "(nine)");
+        return numbers;
+    }
+
+    void presentNumbersAsString() {
+        char[] arr = input.toCharArray();
+        for (char c : arr) {
+            if (fillMap().containsKey(c)) {
```
tromanyuk, line 32: `fillMap` will be executed on every call - this means we're going to add entries to the map every time. This can be done only once during object construction.

java/solomiia_tymoshchuk/src/com/zoolatech/lecture4/tasks/_7/StringRepresentation.java
```diff
@@ -0,0 +1,44 @@
+package com.zoolatech.lecture4.tasks._7;
+
+import java.util.HashMap;
+import java.util.Scanner;
+
+public class StringRepresentation {
+    String input;
```
tromanyuk, line 7: both fields should be defined as private

PR #153, opened 2022-08-17
Cherepnin Oleh: homework for lecture 10 (I/O)

Oleghon:


java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/FileReader.java
```diff
@@ -0,0 +1,43 @@
+package com.zoolatech.lecture10.tasks;
+
+import java.io.*;
+import java.util.Date;
```
ogorbov, line 4: Please avoid using util.Date. Here is why https://programminghints.com/2017/05/still-using-java-util-date-dont/
LocalDateTime will be better

java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/FileReader.java
```diff
@@ -0,0 +1,43 @@
...
+
+public class FileReader {
+    private String path;
+
+    public FileReader(String path) {
+        this.path = path;
+    }
+
+    public void readFile(String name) {
+        StringBuilder builder = new StringBuilder(path);
```
ogorbov, line 14: nit: StringBuilder can be a better option in some cases. Ex concatenation in a loop. If you need to concatenate 2-3 strings + will be better

java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/FileReader.java
```diff
@@ -0,0 +1,43 @@
...
+            try (InputStream stream = new FileInputStream(file)) {
+                String result = new String(stream.readAllBytes());
+                System.out.println(result);
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        }
+    }
+
+    private boolean checkFile(File file) {
```
ogorbov, line 27: nit: The name is not clear what this method checks. let's use isFileExist

java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/Task.java
```diff
@@ -0,0 +1,29 @@
...
+    public static void main(String[] args) {
+        FileReader fileReader = new FileReader("src/com/zoolatech/lecture10/tasks/db/");
+        Scanner scanner = new Scanner(System.in);
+
+        do {
+            System.out.println("Input file name (names, surnames, positions):");
+            String fileName = scanner.next();
+            fileReader.readFile(fileName);
+            System.out.println("Repeat operation (press 'Y' to repeat):");
+        } while (scanner.next().equalsIgnoreCase("Y"));
```
ogorbov, line 27: nit: not about this case but try to use this format "Y".equalsIgnoreCase(scanner.next()). It looks the same but the constant cannot be null. If you use this approach all the time once it will safe you time on fixing null pointer exception

java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/Task.java
```diff
@@ -0,0 +1,30 @@
...
+ * filename to a new file called "error.log" (ideally data should be appended).
+ */
+
+public class Task {
+    public static void main(String[] args) {
+        FileReader fileReader = new FileReader("src/com/zoolatech/lecture10/tasks/db/");
+        Scanner scanner = new Scanner(System.in);
+        String answer = "";
+        do {
+            System.out.println("Input file name (names, surnames, positions):");
```
ogorbov, line 23: Please read all available files from directory using File object instead of hard code

java/Cherepnin_Oleh/src/com/zoolatech/lecture10/tasks/FileReader.java
```diff
@@ -0,0 +1,56 @@
...
+        }
+    }
+
+    public String getFilesAsString() {
+        String collect = Stream.of(new File(path).listFiles())
+                .filter(file -> !file.isDirectory())
+                .map(File::getName)
+                .filter(name -> !name.endsWith("log"))
+                .collect(Collectors.joining(", "));
+        return pattern.matcher(collect).replaceAll("");
```
ogorbov, line 54: nit:
There is a joke about it:
`Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.` - original problem + regular expression. 
In this case, I'd use substring() instead of regex.  It's always hard to deal with regex and try to avoid it if possible =)

ogorbov: @Oleghon just wanted to let you know about this

ogorbov (approved):

PR #154, opened 2022-08-17
[HW10-01] Bogdana Maksiuta homework for lecture 10

danta-m:


java/bogdana_maksiuta/src/com/zoolatech/lecture10/tasks/_1/Main.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+public class Main {
+    static Scanner scanner = new Scanner(System.in);
+
+    public static void main(String[] args) {
+        String directionToTaskFolder = "src/com/zoolatech/lecture10/tasks/_1/";
+        File folder = new File(directionToTaskFolder + "db");
+
+        ArrayList<String> filesList = new ArrayList<>();
```
ogorbov, line 27: nit: List<String> filesList = new ArrayList<>();
I cannot say why it will be better here, but your code depends only on list interface and doesn't care if it's array or linked list implementation.  let's say it's best practice =)

java/bogdana_maksiuta/src/com/zoolatech/lecture10/tasks/_1/Main.java
```diff
@@ -0,0 +1,64 @@
...
+        for (File file : Objects.requireNonNull(folder.listFiles())) {
```
ogorbov, line 28: nit: I like that you added a check here =) I'd check not in a loop. One line before that.
But please pay attention to that it's not a good design in java. To avoid NulllPointerExceptions never return null if the return time is collection. Always return empty collection to avoid such checks on client side

java/bogdana_maksiuta/src/com/zoolatech/lecture10/tasks/_1/Main.java
```diff
@@ -0,0 +1,64 @@
...
+            filesList.add(file.getName());
+            System.out.println(file.getName());
+        }
+
+        System.out.println("Choose the file " + filesList + " :");
+        String userChoice;
+
+        do {
+            userChoice = scanner.next().toLowerCase();
+            if (userChoice.equals("names") ||
```
ogorbov, line 38: Let's check if filename in filesList to avoid having a dependency on some particular name.

java/bogdana_maksiuta/src/com/zoolatech/lecture10/tasks/_1/Main.java
```diff
@@ -0,0 +1,64 @@
...
+                    }
+                } catch (IOException e) {
+                    throw new RuntimeException(e);
+                }
+            } else {
+                try (FileWriter fw = new FileWriter(directionToTaskFolder + "error.log", true);
+                     BufferedWriter bufferedWriter = new BufferedWriter(fw)) {
+                    String errorMessage = "File with name \"" + userChoice + "\" does not exist";
+                    bufferedWriter.write(errorMessage + "\n");
+                    System.out.println(errorMessage);
```
ogorbov, line 56: Please use System.err here

PR #157, opened 2022-08-24
Anton Kozyk HW11

AntonKun:


ogorbov (approved):
I like your implementations. Approved!

Just in case if you are interested in my notes:
- MyAccount and other classes removed init()/destroy() if they are not used
- ServletFilter use status code constant from java
- DatabaseServlet directoryContent can be local variable and add private to resourceURL

However all of this are minor comments that don't worth to request change =)

PR #161, opened 2022-08-30
Cherepnin Oleh: homework for lecture 11

Oleghon:


ogorbov:
I like that you moved further and added some jsp pages. I have a few suggestions but most of them are not that critical
1. LoginFilter - remove init/destroy
2. FileManager you are mixing two concerns here. I suggest moving getBody and convertJson to a separate class like RequestParser
3. FileManager fileName = clearFileName(fileName); - Please don't set a new value to method arguments. It's a very tricky approach
4. FileManager clearFileName -> getFileName
5. FileManager String readFile(String fileName)  -> fileName -> requestUrl
6. DBServlet please use the same name as in WebServlet
7. TableServlet.doPost logging an error in terminal is ok for now, but it's better to notify client somehow. You can return conflict status code in case file exist
8. TableServlet.doPut - 404 when file not founf

Oleghon:
I added changes as per suggestions in comment. 
Check it out, please.

ogorbov:
A few more minor improvements:
1. FileManager I like but let's separate it from web concern. User RequestParser in servlets instead of calling it in FileManager
2. TableServlet/TablesServlet let's move getClassLoader().getResource("dbs/") to FileManager. It will allow us to isolate logic related to files in FileManager. You can init it in FileManager.
3. Rename file_managment -> file
4. When you move db path to FileManager you can add to FileManager createNewFile and updateFille methods. In servlet, you will have the result of operation like boolean value and depending on it return status code and message

Oleghon:
Check out new updates, please.
I mirgated getting of storage path and all CRUD logic in FileManager.
But I couldn`t separate RequestParser from it because the parser is only used when updating a file.

ogorbov:
> Check out new updates, please. I mirgated getting of storage path and all CRUD logic in FileManager. But I couldn`t separate RequestParser from it because the parser is only used when updating a file.

You can parse body in servlet and pass String to FileManger. However, I'll not push this change and will merge pr.

ogorbov (approved):

PR #168, opened 2022-09-03
[JS-HW2] Horlacheva Anna

Horlacheva-Ann:
[https://drive.google.com/file/d/1CMvecb567col3s9lQSdItUqB4mNqvUBm/view?usp=sharing](url)

Link to video screen record 

I ran into several problems. the sound is not connected to the game (maybe I'm trying to connect it in the wrong function) and the timer does not work. These problems are being finalized, I would be very glad to receive an analysis of them or a hint in a solution. Thanks


zoolamyke:
So, the audio won't play because `parcel` server can't return audio files properly.
As a quick workaround, you can launch a regular HTTP server like this: `npx http-server`, and request your assets like this:
image

Horlacheva-Ann:
Thanks! Sound and timer now works correct.

PR #204, opened 2022-10-12
UI skelet

nikson666:


js/chat_app/src/App.js
```diff
@@ -1,23 +1,18 @@
-import logo from './logo.svg';
-import './App.css';
+/* eslint-disable jsx-quotes */
+import React from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import { setStatus } from './redux/slicies/testSlise';
 
 function App() {
+  const dispatch = useDispatch();
+  const status = useSelector((state) => state.test);
```
tretjakan, line 8: I think that if you want to create example - please use something that can make sense

For example, you gonna have authorization - so create "auth" slice for redux-toolkit and leave a simple example where you will be dispatching any action.

js/chat_app/src/App.js
```diff
@@ -1,23 +1,18 @@
...
-          href="https://reactjs.org"
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          Learn React
-        </a>
-      </header>
+    <div>
+      <p>{status.toString()}</p>
+      <button type='submit' onClick={() => dispatch(setStatus())}>
```
tretjakan, line 13: Since we're using MaterialUI, i would ask you to use Buttons from package already

js/chat_app/src/redux/slicies/testSlise.js
```diff
@@ -0,0 +1,16 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+const testSlice = createSlice({
```
tretjakan, line 3: Just make AuthSlice here for example please.

PR #209, opened 2022-10-17
Chat list

NicolasKul:
https://user-images.githubusercontent.com/43201911/198062272-1e86de85-140d-401f-9947-b380215dbe3a.mov

(https://trello.com/c/GdGdB2a0/2-implement-chat-list)
- Added "ChatList.js" component
- Added demo


js/chat_app/src/ChatList.js
```diff
@@ -0,0 +1,171 @@
+import * as React from "react";
```
zoolamyke, line 1:
```suggestion
import React from "react";
```

PR #210, opened 2022-10-18
added  new chat and new user form, and chat list

lilproger21:


js/chat_app/src/App.jsx
```diff
@@ -0,0 +1,34 @@
...
+      </Button>
+      <NewUserForm
+        open={isCreateUserOpen}
+        onClose={() => setIsCreateUserOpen(false)}
+      />
+      <NewChatForm
+        open={isCreateChatOpen}
+        onClose={() => setIsCreateChatOpen(false)}
+      />
+      <Chats />
```
anastasiiaafanasieva, line 29: Maybe place this part in a separate component too?

js/chat_app/src/components/NewUserForm/NewUserForm.jsx
```diff
@@ -0,0 +1,102 @@
...
+        },
+      })
+        .then((response) => response.json())
+        .then((data) => setUsers(data));
+    }, 1000);
+
+    return () => {
+      clearInterval(interval);
+    };
+  }, []);
```
anastasiiaafanasieva, line 35: I've already seen this function in another component (newChatForm), the difference is only in "auth-token", as far as I understood. Maybe we can make a general function helper for requests and pass in fetch two parameters - our URL and also 'options'. 
For example, `
const request = return fetch(URL, options).then(response => response.json())

const options = {
  method: 'POST', // 'GET', 'POST', 'PATCH', 'PUT', 'DELETE' if needed
  headers: {
    'Content-Type': 'application/json,
    'auth-token': adminId or currentUserToken,
  },
};
`

anastasiiaafanasieva (changes requested):
@VikaMartynova @lilproger21 hey! I've added my comments☺️

js/chat_app/src/App.jsx
VikaMartynova: Which exactly component did you mean?
`NewChatForm` and `Chats `?

js/chat_app/src/function/fetch.js
```diff
@@ -0,0 +1,23 @@
+export const getData = async (URL, options = {}) => {
+  return fetch(URL, options).then((response) => response.json());
+};
+
+export const fetchRequestJSON = (URL, method, token, data) => {
```
VikaMartynova, line 5: Nice job!
One reusable function looks much better!

js/chat_app/src/components/NewUserForm/NewUserForm.jsx
```diff
@@ -0,0 +1,83 @@
...
+                  value={userName}
+                  onChange={(e) => setUserName(e.target.value)}
+                />
+                <TextField
+                  label="Password"
+                  type="password"
+                  value={userPassword}
+                  onChange={(e) => setUserPassword(e.target.value)}
+                />
+                <Button fullWidth variant="outlined" onClick={submitHandler}>
```
VikaMartynova, line 72: Here you split submit form logic into two functions. 
Functionally it works but looks weird. 
Could you set the type for the button as submit and then handle all changes in `submitHandler`

js/chat_app/src/components/NewUserForm/NewUserForm.jsx
```diff
@@ -69,7 +66,12 @@ export const NewUserForm = ({ open, onClose }) => {
                   value={userPassword}
                   onChange={(e) => setUserPassword(e.target.value)}
                 />
-                <Button fullWidth variant="outlined" onClick={submitHandler}>
+                <Button
+                  fullWidth
+                  variant="outlined"
+                  onClick={submitHandler}
+                  onSubmit
```
VikaMartynova, line 73: Does this prop mean type="submit" for the button?

js/chat_app/src/components/NewChatForm/NewChatForm.jsx
```diff
@@ -92,7 +92,7 @@ export const NewChatForm = ({ open, onClose }) => {
                     ))}
                 </Select>
               </FormControl>
-              <Button variant="outlined" onClick={submitHandler}>
+              <Button variant="outlined" onClick={submitHandler} onSubmit>
```
VikaMartynova, line 95: From my previous comment, I meant skip onClick for Button and moved all login to onSubmit={submitHandler} for the form tag.

VikaMartynova (changes requested):

VikaMartynova (approved):

js/chat_app/src/App.jsx
anastasiiaafanasieva: I mean to add all these parts including Button, NewUserForm and NewChatForm in a separate component, so in App component we will have Header, LoginPage, separate component with Button, NewUserForm and NewChatForm (it's hard for me to name it in a proper way :)) and Chats. But I'm absolutely ok with the current variant!

anastasiiaafanasieva (approved):

PR #224, opened 2022-11-03
added chat

lilproger21:


js/chat_app/src/App.jsx
```diff
@@ -48,22 +44,12 @@ function App() {
...
-          open={isCreateChatOpen}
-          onClose={() => setIsCreateChatOpen(false)}
-        />
-        <Chats />
+      <div className="container">
+        {userData.authToken && (
+          <>
+            <ButtonsAndChatList userData={userData} />
+          </>
+        )}
```
VikaMartynova: `ThemeProvider`  should work for the whole project; please, add this block inside it.

js/chat_app/src/components/Chat/Chat.jsx
```diff
@@ -0,0 +1,102 @@
...
+  useEffect(() => {
+    if (chatId) {
+      fetchRequestJSON("/api/users", "GET", userData.authToken).then((data) =>
+        setUsersList(data)
+      );
+      fetchRequestJSON(`/api/chats/${chatId}`, "GET", userData.authToken).then(
+        (data) => setMessagesList(data.messages)
+      );
+      const interval = setInterval(() => {
+        if (chatId !== "") {
```
VikaMartynova: Why do you need this condition?

js/chat_app/src/components/Chat/Chat.jsx
```diff
@@ -0,0 +1,102 @@
...
+          alignItems: "center",
+          justifyContent: "space-around",
+        }}
+      >
+        <div>Click on chat</div>
+      </Box>
+    );
+  } else {
+    return (
+      <Container>
```
VikaMartynova, line 61: Do you need this `Container`? Looks like it only covers one tag, `Box`.

js/chat_app/src/components/NewChatForm/NewChatForm.jsx
```diff
@@ -14,11 +14,9 @@ import {
 } from "@mui/material";
 import { fetchRequestJSON } from "../../function/fetch";
 
-const currentUserToken = "d46c97b0-ddcc-4897-b327-8593fcf6cb78";
-const currentUserId = "75da786f-b490-408d-9ea1-ee1675b98d3c";
 const intervalUpdate = 1000;
 
-export const NewChatForm = ({ open, onClose }) => {
+export const NewChatForm = ({ open, onClose, userInfo }) => {
```
VikaMartynova: If, in some way, `userInfo` equals null or undefined, you will get an error, so you can, at least, add the default value `userInfo = {}` but here you have `userInfo.user.id` so you will get an error in this case too; or you can add checks for property existing `userInfo?.user?.id`.

js/chat_app/src/components/NewUserForm/NewUserForm.jsx
```diff
@@ -11,25 +11,26 @@ import {
 import { UsersList } from "../UsersList/UsersList";
 import { fetchRequestJSON } from "../../function/fetch";
 
-const adminId = "c8e26274-93ee-4acb-9f51-126264adaeb2";
 const intervalUpdate = 1000;
 
-export const NewUserForm = ({ open, onClose }) => {
+export const NewUserForm = ({ open, onClose, userInfo }) => {
```
VikaMartynova: The same is here

VikaMartynova (changes requested):

VikaMartynova (approved):

PR #225, opened 2022-11-03
Return working code without creation chat form

leonid-06:
The situation is as follows:
There was a working code, and along with it there were roots to add their changes to the master branch. And then one inexperienced programmer mistakenly made the wrong changes to the main branch and broke everything.
But now he has drawn conclusions and fixed everything


js/chat_app/package.json
```diff
@@ -56,4 +56,4 @@
     "nodemon": "^2.0.20",
```
miromyke, line 59: Please, delete `package-lock.json`, and revert `yarn.lock` if the change wasn't intentional.

PR #226, opened 2022-11-03
Implement chat members list

nikson666:


js/chat_app/src/modals/ChatForm/Select.jsx
```diff
@@ -0,0 +1,61 @@
...
+import TextField from '@mui/material/TextField';
+import MenuItem from '@mui/material/MenuItem';
+import FormControl from '@mui/material/FormControl';
+import { useDispatch, useSelector } from 'react-redux';
+import { setSelectedMembers } from '../../redux/chat/slice';
+import authSelectors from '../../redux/auth/selector';
+import usersSelectors from '../../redux/users/selector';
+
+export default function CreateChatSelect({ resetMembersTrigger }) {
+  const [membersName, setMembersName] = useState([]);
```
tretjakan: for arrays try to use the plural always. like `memberNames`

js/chat_app/src/modals/ChatMembersForm/MemberList.jsx
```diff
@@ -0,0 +1,46 @@
...
+import ListItemAvatar from '@mui/material/ListItemAvatar';
+import ListItemText from '@mui/material/ListItemText';
+import DialogTitle from '@mui/material/DialogTitle';
+import Dialog from '@mui/material/Dialog';
+import DialogContent from '@mui/material/DialogContent';
+import MemberAvatar from './MemberAvatar';
+import Select from './SelectForm';
+
+function MemberList({ onClose, open, membersData, allUsers }) {
+  const usersDontMembers = allUsers.filter((user) => !membersData?.includes(user));
```
tretjakan: just use `users` there instead of `usersDontMembers`

js/chat_app/src/modals/ChatMembersForm/MemberList.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <Dialog onClose={onClose} open={open}>
+      <DialogTitle>Group members</DialogTitle>
+      <DialogContent
+        sx={{
+          width: 450
+        }}
+      >
+        <Select usersDontMembers={usersDontMembers} />
```
tretjakan: the same as above

js/chat_app/src/modals/ChatMembersForm/SelectForm.jsx
```diff
@@ -0,0 +1,91 @@
...
+import MenuItem from '@mui/material/MenuItem';
+import { useDispatch, useSelector } from 'react-redux';
+import { addChatMembersThunk } from '../../redux/chat/slice';
+import chatSelectors from '../../redux/chat/selector';
+
+function Select({ usersDontMembers }) {
+  const [selectedMembers, setSelectedMembers] = useState([]);
+  const [selectesMembersError, setSelectesMembersError] = useState(false);
+
+  const [membersId, setMembersId] = useState([]);
```
tretjakan: also try to use plural, like `memberIds`

js/chat_app/src/redux/chat/slice.js
```diff
@@ -0,0 +1,99 @@
...
+    const response = await chatServices.addChatMembers(chatId, authToken, members);
+    dispatch(getChatByIdThunk(response?.id, authToken));
+  }
+);
+
+export const createChatThunk = createAsyncThunk(
+  'createChat/api/chats',
+  async (title, { getState, dispatch, rejectWithValue }) => {
+    try {
+      const { authToken } = getState().auth.auth;
```
tretjakan: use selectors there too

js/chat_app/src/redux/chat/slice.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      return rejectWithValue(error.response.data);
+    }
+  }
+);
+
+export const getChatByIdThunk = createAsyncThunk('getChatById/api/chats/:id', async (chatId, { getState }) => {
+  try {
+    const { authToken } = getState().auth.auth;
```
tretjakan: use selectors there too

js/chat_app/src/services/chatServices.js
```diff
@@ -0,0 +1,22 @@
...
+      title
+    };
+    const response = await query('/api/chats', 'POST', authToken, body);
+    return response;
+  },
+  addChatMembers: async (chatId, authToken, members) => {
+    const body = {
+      members
+    };
+    const response = query(`/api/chats/${chatId}/members`, 'POST', authToken, body);
```
tretjakan: why without `await` here?

just `return query('/api/chats/${chatId}/members', 'POST', authToken, body);` instead.

js/chat_app/src/services/utils/utils.js
```diff
@@ -0,0 +1,27 @@
+export const query = async (url, method, authToken, body = {}) => {
+  try {
+    if (Object.keys(body).length) {
+      const data = await fetch(url, {
+        method,
+        headers: {
```
tretjakan: you can make headers as separate object

like
`
const headers = {
          'content-type': 'application/json',
          'Auth-Token': authToken
}
`

js/chat_app/src/redux/chat/slice.js
nikson666: getState() doesn't work with selectors for useSelector()
getState() gives us access to the shared storage, after which we can iterate over the individual state keys and get data from them;
image

js/chat_app/src/redux/chat/slice.js
nikson666: the same as above

js/chat_app/src/redux/chat/slice.js
tretjakan: so what? :) You can do something like

`const getAuth = (state) => state.auth.auth`;

And call it like
`getAuth(getState())`

